PR #153, opened 2018-01-13
tools: build propel_windows with package.js

piscisaureus:
Also includes various other fixes to make package.js work on windows,
as well as some stylistic clean-ups.

### DIFF
--- a/tools/package.js
+++ b/tools/package.js
@@ -6,12 +6,12 @@ const run = require("./run");
 const fs = require("fs");
 const path = require("path");
 const config = require("./config");
-const {execSync} = require("child_process");
+const { execSync } = require("child_process");
 
 const clean = process.argv.includes("clean");
 
 if (clean) {
-  run.rmrf("build")
+  run.rmrf("build");
 }
 
 const v = run.version();
@@ -19,7 +19,6 @@ console.log("version", v);
 
 // Build the binding.
 run.sh("node tools/build_binding.js");
-
 
 function createPackageJson(src, dst, packageJson = {}) {
   let p = JSON.parse(fs.readFileSync(src, "utf8"));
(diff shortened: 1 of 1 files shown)


tools/package.js
```diff
@@ -35,55 +34,60 @@ function createPackageJson(src, dst, packageJson = {}) {
...
-                    distDir + "/tensorflow-binding.node");
-    fs.copyFileSync("build/Release/libtensorflow.so",
-                    distDir + "/libtensorflow.so");
-    fs.copyFileSync("build/Release/libtensorflow_framework.so",
-                    distDir + "/libtensorflow_framework.so");
+                    distDir + "/tensorflow-binding.node"
+    );
+    if (process.platform === "win32") {
+      fs.copyFileSync("build/Release/tensorflow.dll",
+                      distDir + "/tensorflow.dll");
```
ry: Does it work without libtensorflow_framework.so?

tools/package.js
```diff
@@ -98,8 +102,8 @@ function buildAndTest() {
                     path.join(testDir, "package.json"));
 
   process.chdir(testDir);
-  run.sh("npm install " + propelPkgFn);
-  run.sh("npm install " + tfPkgFn);
+  execSync("npm install " + propelPkgFn, { stdio: "inherit" });
+  execSync("npm install " + tfPkgFn, { stdio: "inherit" });
```
ry: What's this change about?

tools/package.js
piscisaureus: That doesn't exist on windows. All you need is tensorflow.dll.

tools/package.js
piscisaureus: npm is a .bat file.
spawnSync cannot spawn bat files.

ry: k

PR #160, opened 2018-01-15
Website edits

ry:


website/style.css
```diff
@@ -315,6 +325,29 @@ header h1 {
...
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  padding: 16px;
+  height: 32px;
+  margin: 0 auto;
+}
+
+.home-link {
+  background: url('propelLogo.svg');
```
piscisaureus: Landed it for now. I don't like case sensitive filenames though, would rather name this propel_logo or propel-logo.

ry: I removed .home-link since it's not being used right now - and propelLogo.svg is not checked in. cc @yungbeto

PR #163, opened 2018-01-18
fork dl

piscisaureus:


jasmine_shim.ts
```diff
@@ -0,0 +1,221 @@
...
+        for (const key in { ...a, ...b }) {
+          if (!compare(a[key], b[key])) {
+            return false;
+          }
+        }
+        seen.set(a, b);
+        return true;
+      }
+      return false;
+    })(value, expected);
```
ry: This looks useful. Maybe this logic should be in util.ts? Our current equality doesn't handle strings.

jasmine_shim.ts
```diff
@@ -0,0 +1,221 @@
...
+      }
+      if (isNaN(a) && isNaN(b)) {
+        return true;
+      }
+      if (a && typeof a === "object" && b && typeof b === "object") {
+        if (seen.get(a) === b) {
+          return true;
+        }
+        for (const key in { ...a, ...b }) {
+          if (!compare(a[key], b[key])) {
```
ry: impressive

util.ts
```diff
@@ -20,6 +20,7 @@ const debug = false;
 const J = JSON.stringify;
 export const IS_WEB = typeof window !== "undefined";
 export const IS_NODE = !IS_WEB;
+export const global = (0, eval)(IS_WEB ? "window" : "global");
```
ry: Can you avoid this comma operator? It's rather obtuse.

piscisaureus: done

jasmine_shim.ts
piscisaureus: This can be done later.

test_browser.ts
```diff
@@ -36,7 +36,7 @@ const TESTS = [
   { href: "notebook_mnist.html" }
 ];
 
-const debug = !!process.env.PP_DEBUG;
+const debug = !!process.env.PP_TEST_DEBUG;
```
ry: Can you add a comment explaining what this does

test_browser.ts
```diff
@@ -37,6 +37,15 @@ const TESTS = [
 ];
 
 const debug = !!process.env.PP_TEST_DEBUG;
+const testdl = !!process.env.PP_TEST_DL;
```
ry: Maybe document both of these env vars at the top of the file?

test_browser.ts
piscisaureus: done

PR #167, opened 2018-01-20
Huge website refactor to use React and cache notebook output

ry:
Known Issues:
- New cell button in notebook.html doesn't work.
- Shift-Enter doesn't change focus to the next cell.
- `notebook_mnist.html` was removed (this is ok - we can bring it back once memory management is done)
- syntax highlighting was removed for the "npm install" examples on `index.html` (this is okay for now)
- `notebook-cell-running` and `notebook-cell-updating` are not working.


nb_transpiler.ts
```diff
@@ -18,6 +18,7 @@ import { assert } from "./util";
 
 const importFn = "__import";
 const globalVar = "__global";
+const consoleVar = "__console";
```
piscisaureus: this variable is not used.

tools/build_website.ts
```diff
@@ -0,0 +1,80 @@
...
+        }
+        const bodyHtml = document.body.innerHTML;
+        const html =  website.getHTML(page.title, bodyHtml);
+        resolve(html);
+      } catch (e) {
+        reject(e);
+      }
+    });
+  });
+  return p;
```
piscisaureus: return await p?

nb_transpiler.ts
ry: fixed

tools/build_website.ts
ry: I think it's correct as is

ry:
Will do

PR #168, opened 2018-01-20
ci: add appveyor (Windows CI)

piscisaureus:


.appveyor.yml
```diff
@@ -0,0 +1,27 @@
+version: "{branch}-build{build}"
+
+environment:
+  matrix:
+    - appveyor_build_worker_image: Visual Studio 2015
+      nodejs_version: 8
+
+    - appveyor_build_worker_image: Visual Studio 2017
+      nodejs_version: 9
```
ry: Does having both double the presubmit time?

piscisaureus: Yeah, it is kinda slow. I'll remove the node v9 build from the matrix

PR #171, opened 2018-01-21
Adds scss support to website 

ry:
(reduces number of requests to load style sheets from 5 to 1)


website/main.scss
```diff
@@ -1,5 +1,10 @@
+@import './normalize.scss';
+@import './skeleton.scss';
+@import './codemirror.scss';
+@import './syntax.scss';
 @import url('https://fonts.googleapis.com/css?family=Roboto|Roboto+Mono');
 
+
```
piscisaureus: ...?

piscisaureus:
why is normalize.css duplicated, while the others are renamed?

tools/build_website.ts
```diff
@@ -62,13 +63,26 @@ async function writePages() {
   }
 }
 
+function scss(inFile, outFile) {
+  const options = {
+    file: inFile,
+    includePaths: ["./website"],
+  };
+
+  const result = renderSync(options).css.toString("ascii");
```
piscisaureus: why ascii and not utf8?

website/main.scss
ry: Fixed.

tools/build_website.ts
ry: Fixed.

ry:
normalized fixed too.

piscisaureus:
lgtm

PR #173, opened 2018-01-22
Fix run button, new cell button, and stylelint

ry:


piscisaureus:
LGTM
please rebase to make the tests pass
note that shift+enter behavior is still broken

PR #177, opened 2018-01-23
src: merge tsconfig.json with dl/tsconfig.json

piscisaureus:
* Strict rules that used to apply only to the DeepLearn source code are
  now applied everywhere.
* With one exception: the "noImplicitAny" rule is off.
* This commit does not touch tslint rules.


api_test.ts
```diff
@@ -399,7 +399,7 @@ test(async function api_matMul() {
 });
 
 test(async function api_reduceSum() {
-  for (const [device, $] of deviceTests()) {
+  for (const [, $] of deviceTests()) {
```
ry: this seems to be an error?

nn_example.ts
```diff
@@ -20,7 +20,6 @@ import * as plt from "./matplotlib";
 import * as mnist from "./mnist";
 import { assert, IS_WEB } from "./util";
 
-const debug = true;
```
ry: removed because of unused variable?

piscisaureus: Yeah

piscisaureus:
PTAL.

The GPU tests are actually failing. That's not due to this PR though; the errors are:

```
> No data found for NDArray with data id 1223. Use dl.ENV.math instead of constructing your own NDArrayMath. If you need to construct your own math, make sure this array is allocated after the math construction
```

and 

```
> Unknown data type uint8
```

api_test.ts
```diff
@@ -43,12 +43,13 @@ function gpuConvert(x: TensorLike, args?: types.TensorOpts): Tensor {
 }
 
 // Allows tests to run on CPU:0 and GPU:0 (if available).
-function deviceTests(): Array<[string, ConvertFn]> {
-  const out: Array<[string, ConvertFn]> = [ ["CPU:0", $] ];
+function testDevices(
+  fn: ($: ConvertFn, device: string) => Promise<void>
+): void {
+  test({ fn: () => fn($, "CPU:0"), name: `${fn.name} CPU:0` });
```
ry: Nice

ry:
This branch (da816a6f9b87f671ae31fcd373c544ee2722a080) passes tests on the CUDA machine.

api_test.ts
```diff
@@ -43,12 +43,13 @@ function gpuConvert(x: TensorLike, args?: types.TensorOpts): Tensor {
...
   if (gpuAvail()) {
-    out.push(["GPU:0", gpuConvert]);
+    test({ fn: () => fn(gpuConvert, "GPU:0"), name: `${fn.name} GPU:0` });
```
ry: Can you change the name to fn.name + "_GPU" and similarly above for CPU


(Spaces don't work well for filtering.. I was trying to go to "http://127.0.0.1:8081/static/test.html#script=/test_isomorphic.js&filter=reduceMean GPU:0")

piscisaureus: did that

PR #187, opened 2018-01-26
Implements tensor.setDiag and eye

ry:
- [X] cuda tests checked
- [X] webgl test checked

### DIFF
--- a/api_test.ts
+++ b/api_test.ts
@@ -1023,6 +1023,27 @@ testDevices(async function api_neuralNet(T, device) {
   }
 });
 
+testDevices(async function api_setDiag(T, device) {
+  const matrix = T(zeros([4, 3]));
+  const m = matrix.setDiag([1, 2, 3]);
+  assertAllEqual(m, [
+    [1, 0, 0],
+    [0, 2, 0],
+    [0, 0, 3],
+    [0, 0, 0]
+  ]);
+});
+
+testDevices(async function api_eye(T, device) {
+  const eye4 = T(api.eye(4));
+  assertAllEqual(eye4, [
+    [1, 0, 0, 0],
+    [0, 1, 0, 0],
+    [0, 0, 1, 0],
+    [0, 0, 0, 1]
+  ]);
+});
+
 if (IS_NODE) {
   test(async function api_inspect() {
     const t = T([1, 2, 3]);
--- a/dl.ts
+++ b/dl.ts
@@ -25,7 +25,7 @@ import { MathBackendWebGL }
 import { MatrixOrientation }
   from "./dl/math/backends/types/matmul";
 import { NDArrayMath } from "./dl/math/math";
-import { Array2D, Array3D, Array4D, NDArray, Scalar }
+import { Array1D, Array2D, Array3D, Array4D, NDArray, Scalar }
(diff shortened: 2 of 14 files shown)


dl.ts
```diff
@@ -191,8 +191,10 @@ export class OpsDL implements types.BackendOps {
     return new TensorDL(ndarray, x.math);
   }
 
-  eye(size: number, dtype: types.DType = "float32"): types.BasicTensor {
-    throw new Error("Not Implemented");
+  setDiag(x: TensorDL, diag: TensorDL): TensorDL {
+    ENV.setMath(x.math);
+    const nd = x.math.setDiag(x.ndarray as Array2D, diag.ndarray as Array1D);
```
piscisaureus: What happens when `x` or `diag` don't have the expected rank?

ops.ts
```diff
@@ -465,3 +465,8 @@ export const oneHot = defFW("oneHot",
     return bo.oneHot(x, depth, onValue, offValue);
   });
 defBW("oneHot", null);
+
+export const setDiag = defFW("setDiag", (input, diag) => {
+  return bo.setDiag(input, diag);
+});
+defBW("setDiag", null);  // TODO
```
piscisaureus: Are you going to land this without a backwards pass?

api_test.ts
```diff
@@ -1023,6 +1023,27 @@ testDevices(async function api_neuralNet(T, device) {
   }
 });
 
+testDevices(async function api_setDiag(T, device) {
```
piscisaureus: You are not using the `device` parameter, so the GPU backend isn't stressed.

api_test.ts
```diff
@@ -1023,6 +1023,27 @@ testDevices(async function api_neuralNet(T, device) {
...
+  const m = matrix.setDiag([1, 2, 3]);
+  assertAllEqual(m, [
+    [1, 0, 0],
+    [0, 2, 0],
+    [0, 0, 3],
+    [0, 0, 0]
+  ]);
+});
+
+testDevices(async function api_eye(T, device) {
```
piscisaureus: Ditto (device not used).

dl.ts
ry: It probably blows up. I'd like to leave this as undefined behavior for now, because the TF side [supports arbitrarily ranked inputs](https://github.com/tensorflow/tensorflow/blob/86c10063c8521bd5482df13676bac3575540d9e2/tensorflow/core/ops/ops.pbtxt#L12475-L12493) and I suspect we would want to also support arbitrary rank tensors on the DL side too. So I would put this off until someone needs it and just let it blow up now.

piscisaureus:
Can't you just add `if (x.shape.length !== 2 || diag.shape.length !== 1)
throw new Error("Not supported")` ?

dl.ts
ry: Done

ops.ts
ry: I made it throw not implemented.

api_test.ts
ry: Fixed.

api_test.ts
ry: Fixed.

PR #189, opened 2018-01-27
Replace webpack with parcel

ry:


ry:
What's the right way to create a temp dir in windows?

piscisaureus:
* `tmproot = process.platform === "win32" ? process.env.TEMP : "/tmp"`
* add a script to just run parcel without running build
* Are you going to remove webpack from deps?

package.json
```diff
@@ -21,10 +21,10 @@
     "acorn": "^5.3.0",
     "codemirror": "^5.33.0",
     "d3": "^4.12.2",
-    "he": "^1.1.1",
```
piscisaureus: The `he` dependency is required for gendoc.

ry: No longer, now that react generates the html.

ry:
webpack is removed from deps in last commit

PR #193, opened 2018-01-30
Modify notebook output caching behavior.

ry:
Previously nb output was cached in a server-side render, and
used in the browser without re-executing the cell. Now output
is cached server side, but re-executed client side (only if the user has
JS enabled, obviously).

This is to remove the hack where innerHTML of output cells was
pulled during browser-side initial page render. This optimization
is complicating development.

This also re-enables the notebook-cell-updating feature.

### DIFF
--- a/util.ts
+++ b/util.ts
@@ -240,3 +240,10 @@ export function inferShape(arr: number | boolean | RegularArray<number> |
   }
   return shape;
 }
+
+// Like setTimeout but with a promise.
+export function delay(t: number): Promise<void> {
+  return new Promise(function(resolve) {
+    setTimeout(resolve, t);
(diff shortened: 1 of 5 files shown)


util.ts
```diff
@@ -240,3 +240,10 @@ export function inferShape(arr: number | boolean | RegularArray<number> |
   }
   return shape;
 }
+
+// Like setTimeout but with a promise.
+export function delay(t: number): Promise<void> {
+  return new Promise(function(resolve) {
+    setTimeout(resolve.bind(null), t);
```
piscisaureus: why bind?

ry: Removed.

PR #200, opened 2018-01-30
Fix DL memory leaks

piscisaureus:


dl/math/backends/backend_cpu.ts
```diff
@@ -32,7 +32,7 @@ import {MathBackend} from './backend';
 import {MatrixOrientation} from './types/matmul';
 
 export class MathBackendCPU implements MathBackend {
-  private data: {[dataId: number]: DataTypeMap[DataType]} = {};
+  private dataMap: WeakMap<Symbol, DataTypeMap[DataType]> = new WeakMap();
```
ry: s/Symbol/DataId/ ?

dl/math/backends/backend_cpu.ts
```diff
@@ -41,18 +41,18 @@ export class MathBackendCPU implements MathBackend {
...
     if (values == null) {
       throw new Error('MathBackendCPU.write(): values can not be null');
     }
     this.throwIfNoData(dataId);
-    this.data[dataId] = values;
+    this.dataMap.set(dataId, values);
   }
   writePixels(
-      dataId: number,
+      dataId: Symbol,
```
ry: DataId or Symbol?

dl/math/backends/backend_cpu.ts
piscisaureus: Also note thatn eventually this should all go away - the texture should just be a property on the NDArray, rather than having this lookup table (and the "registeredArrays" refcount map).

dl/math/ndarray.ts
```diff
@@ -62,17 +62,18 @@ export interface ShapeMap {
   higher: number[];
 }
 
+export class DataId {};
```
ry: Why not use numbers? (I imagine the answer is either because WeakMap doesn't support it, or because incrementing numbers is more error prone)

piscisaureus: Conceptually it wouldn't work. The advantage of WeakMap is that an element is removed from the map when it's key becomes unreferenceable. But numbers never become unreferenceable, so the weakmap would just act as a normal map.
This is all irrelevant though - WeakMap keys aren't allowed to be primitives.

dl/math/backends/webgl/texture_manager.ts
```diff
@@ -82,11 +80,12 @@ export class TextureManager {
   }
 
   dispose() {
-    this.allocatedTextures.forEach(texture => {
-      this.gpgpu.deleteMatrixTexture(texture);
-    });
+    for (const shapeKey in Object.keys(this.freeTextures)) {
+      for (const texture in this.freeTextures[shapeKey]) {
```
ry: s/in/of/ ?

dl/math/backends/webgl/texture_manager.ts
```diff
@@ -82,11 +80,12 @@ export class TextureManager {
...
+        this.gpgpu.deleteMatrixTexture(texture);
+      }
+    }
     this.freeTextures = null;
```
ry: Were the freeTextures not being cleaned up before? Is this a bug in DL?

piscisaureus: Previously the texture manager kept a separate array of *all* allocated textures (`allocatedTextures`) -- which would include the ones listed in freeTextures. I removed that because it was keeping textures alive that would otherwise be garbage collectible. That's why I now loop only the list-of-lists of free textures.

dl/math/backends/webgl/texture_manager.ts
piscisaureus: Two times. Good catch. Fixed.

PR #203, opened 2018-01-31
Write nextCell

Qti3e:
Hi
I've done a TODO

### DIFF
--- a/notebook.ts
+++ b/notebook.ts
@@ -143,7 +143,7 @@ export class Cell extends Component<Props, State> {
       this.editor.on("blur", this.blur.bind(this));
       this.editor.setOption("extraKeys", {
         "Ctrl-Enter": () =>  { this.update(); return true; },
-        "Shift-Enter": () => { this.update(); this.nextCell(); return true; }
+        "Shift-Enter": () => { this.update(); this.focusNextCell(); return true; }
       });
     }
     // Execute the cell automatically.
@@ -164,7 +164,7 @@ export class Cell extends Component<Props, State> {
     }
   }
 
-  async nextCell() {
+  async focusNextCell() {
     const currentId = "cell" + this.id;
     const noteBooks = document.querySelectorAll(".notebook-cell");
     let found = false, i;


Qti3e:
Sorry for the test results :|
Gonna fix it ASAP

notebook.ts
```diff
@@ -164,8 +164,35 @@ export class Cell extends Component<Props, State> {
...
+    }
+    let cell = noteBooks[i];
+    if (!cell) {
+      // create a new cell
+      const newBtn = document.querySelector("#newCell");
+      if (!newBtn) {
+        return;
+      }
+      newBtn.click();
+      await delay(100);
```
ry: Why the delay?

notebook.ts
```diff
@@ -164,8 +164,35 @@ export class Cell extends Component<Props, State> {
...
   }
 
-  nextCell() {
-    // TODO
+  async nextCell() {
+    const currentId = this.id;
+    const noteBooks = document.querySelectorAll(".notebook-cell");
+    let found = false, i;
+    for (i = 0; i < noteBooks.length; i++) {
+      if (noteBooks[i].id === currentId) {
```
ry: When I run `./tools/build_website.ts` it gives me some errors
```
notebook.ts (168,23): Type 'number' cannot be converted to type 'string'. (2352)
notebook.ts (188,14): Property 'click' does not exist on type 'Element'. (2339)
notebook.ts (195,57): Property 'focus' does not exist on type 'Element'. (2339)
```
This actually shows there's a hole in our tests - it shouldn't be green. I will fix this.

ry: Disregard this comment. It seems to be ok...

notebook.ts
Qti3e: Run this code in your console at [Notebook](http://propelml.org/notebook/)
```js
let o = document.querySelectorAll('.notebook-cell')
$('#newCell').click()
console.log(document.querySelectorAll('.notebook-cell').length == o.length)
setTimeout(function(){
    console.log(document.querySelectorAll('.notebook-cell').length == o.length)
}, 100)
```
Result:
```
true
false
```

It's simply because rendering to DOM is not synchronous

Qti3e:
What about scrolling animation?
Which package do you prefer to use?
Personally I think [scroll-to-element](https://www.npmjs.com/package/scroll-to-element) is a good choice

notebook.ts
```diff
@@ -164,8 +164,35 @@ export class Cell extends Component<Props, State> {
     }
   }
 
-  nextCell() {
-    // TODO
+  async nextCell() {
+    const currentId = this.id;
```
ry: this.id is a number.

notebook.ts
ry: currentId is a number, but notebooks[i].id is a string like "cell1" - so I don't think this works. In fact, I'm surprised and concerned that typescript isn't complaining about the mismatched types here.

Qti3e:
Fixed,
Ok no problem also I started to write another local Notebook (Something like Jupyter) with React, publish the beta next week after doing some works :)

notebook.ts
```diff
@@ -164,8 +164,35 @@ export class Cell extends Component<Props, State> {
...
+        break;
+      }
+    }
+    if (!found) {
+      return;
+    }
+    let cell = noteBooks[i];
+    if (!cell) {
+      // create a new cell
+      const newBtn = document.querySelector("#newCell");
```
ry: Rather than trying to find the new button, can you extra id of the cell (`Number(cell.id.replace("cell", ""))`) and use `lookupCell()` to grab the actual Component and then call `.focus` on it?

notebook.ts
```diff
@@ -164,8 +164,35 @@ export class Cell extends Component<Props, State> {
     }
   }
 
-  nextCell() {
-    // TODO
+  async nextCell() {
```
ry: rename to focusNextCell

notebook.ts

notebook.ts
Qti3e: Renaming it on CodeMirror's Shift+Enter binding cause "This branch has conflicts that must be resolved."  and if I don't rename it in that place it would generate a ts type error

ry: Have you rebased your branch?

Qti3e:
I close it because my local branch is shitty and I couldn't make things right
An straight & clean PR is on the way with all your requests

PR #207, opened 2018-02-01
Implement server-side rendering for imshow

### LINKED ISSUE #199: Add docs for imshow
jsdom doesn't support canvas, so server-side rendering is hard. example code:

```
import { randn, imshow } from "propel"
img = randn([300, 300]).relu().mul(255);
imshow(img)
```

Qti3e:
Fixes #199

### DIFF
--- a/matplotlib.ts
+++ b/matplotlib.ts
@@ -12,9 +12,10 @@
    See the License for the specific language governing permissions and
    limitations under the License.
  */
+import * as Canvas from "canvas";
 import * as d3 from "d3";
 import { Tensor } from "./api";
-import { assertEqual } from "./util";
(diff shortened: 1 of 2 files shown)


Qti3e:
Oh :(
How to add a new dependency to Propel?
Does it require to change [propelml/propel_deps](https://github.com/propelml/propel_deps)? :thinking:

ry:
Thanks for looking into this. But I don't feel comfortable taking node-canvas as a dependency at this time - it's very large and has a lot of native code. I think a good short term solution for #199 would be to skip server-side rendering of canvas elements - and only render client side.

Qti3e:
OK :)

PR #208, opened 2018-02-01
Enable typeCheck in ts-node by default

ry:
For unknown reasons ts-node 4.1.0 doesn't report typescript errors like previous versions did. For example the code
```javascript
if (5 === "5") console.log("what");
```
didn't trigger an error with ts-node 4.1.0 but does with 3.3.0. We don't particularly care about ts-node - so downgrading is the easiest fix.

This popped up while reviewing #203. Thanks Parsa.


ry:
I un-did the downgrade and enabled the type check flag.

PR #211, opened 2018-02-02
dl: simplifications

piscisaureus:


dl/math/backends/backend_engine.ts
```diff
@@ -105,15 +105,7 @@ export class BackendEngine {
...
-    if (this.scopeStack.length === 1) {
-      if (this.safeMode) {
-        throw new Error(
-            'You are using math in safe mode. Enclose all ' +
-            'math.method() calls inside a scope: ' +
-            'math.scope(() => {math.method();...}) to avoid memory ' +
-            'leaks.');
-      }
-    } else {
+    if (this.scopeStack.length > 1) {
```
ry: Let me make sure I get this: safeMode used to warn when you created a NDArray outside any scope. Because of the memory leak fix in the previous PR, global NDArrays are properly cleaned up, thus this is not needed?

dl/math/math.ts
```diff
@@ -278,8 +277,7 @@ export class NDArrayMath implements NDArrayManager {
         x.size === util.sizeFromShape(newShape),
         'new shape and old shape must have the same number of elements.');
 
-    return this.backendEngine.executeKernel(
-               'Reshape', {inputs: {x}, args: {newShape}}) as T;
+    return NDArray.make(newShape, {dataId: x.dataId}, x.dtype) as T;
```
ry: Just checking - this passes DL's tests?

dl/math/math.ts
```diff
@@ -288,8 +286,17 @@ export class NDArrayMath implements NDArrayManager {
    */
   cast<D extends DataType, R extends Rank>(
       x: NDArray<DataType, R>, newDType: D): RankMap<D>[R] {
-    return this.backendEngine.executeKernel(
-               'Cast', {inputs: {x}, args: {newDType}}) as RankMap<D>[R];
+    if (!util.hasEncodingLoss(x.dtype, newDType)) {
```
ry: What does hasEncodingLoss do?

dl/math/math.ts
```diff
@@ -2114,9 +2085,8 @@ export class NDArrayMath implements NDArrayManager {
               `but got rank ${offset.rank}.`);
     }
 
-    return this.backendEngine.executeKernel(
-        'BatchNorm4D',
-        {inputs: {x, mean, variance, scale, offset}, args: {varianceEpsilon}});
+    return this.backend.batchNormalization4D(x, mean, variance,
+      varianceEpsilon, scale, offset);
```
ry: This comment is for the whole commit where you switch backendEngine for direct backend access: Nice!

dl/math/backends/backend_engine.ts
```diff
@@ -19,8 +19,6 @@ import * as util from '../../util';
 import {DataType, NDArray} from '../ndarray';
 
 import {MathBackend} from './backend';
-import * as kernel_registry from './kernel_registry';
-import {KernelConfigRegistry} from './kernel_registry';
```
ry: This comment is for the whole commit "remove kernel registry, executeKernel, and related types"
Very nice. I love removing code.

dl/math/math.ts
```diff
@@ -92,7 +92,7 @@ export class NDArrayMath implements NDArrayManager {
       this.customBackend = true;
       this.backend = backend;
     }
-    this.backendEngine = new BackendEngine(this.backend);
+    this.backendEngine = new BackendEngine();
```
ry: Sorry for the un-constructive comments - but i love all these clean ups. It's hard for me to really grok the implications, but it looks very nice. Generally with this PR, what's your greatest concern about removing this code?

dl/environment.ts
```diff
@@ -28,9 +27,6 @@ export enum Type {
 export interface Features {
   // Whether the disjoint_query_timer extension is an available extension.
   'WEBGL_DISJOINT_QUERY_TIMER_EXTENSION_ENABLED'?: boolean;
-  // Whether the timer object from the disjoint_query_timer extension gives
-  // timing information that is reliable.
-  'WEBGL_DISJOINT_QUERY_TIMER_EXTENSION_RELIABLE'?: boolean;
```
ry: Why is this being removed? Do we not care about the timer object being reliable? Is it always reliable?

dl/math/backends/backend_engine.ts
piscisaureus: I am actual slightly doubtful about removing safeMode, but I don't think it has a lot of value for us either.

piscisaureus: Let's discuss on our call.

dl/math/math.ts
piscisaureus: yep

dl/environment.ts
piscisaureus: This 'feature flag' is being set but it is never read.
It's also not super spectacular, it's value is basically `(WEBGL_DISJOINT_QUERY_TIMER_EXTENSION_ENABLED && !isMobileDevice())`.
The `WEBGL_DISJOINT_QUERY_TIMER_EXTENSION_ENABLED` is used at a few places, and not all of them make sense.

dl/math/math.ts
```diff
@@ -566,7 +535,7 @@ export class NDArrayMath implements NDArrayManager {
     const reduceShape = shapes[1];
     const reduceSize = util.sizeFromShape(reduceShape);
     return this.executeOp('mean', () => {
-      return this.scope(keep => {
+      return this.scope(() => {
```
ry: the only way to keep things now is to return them?

piscisaureus: Yes.

dl/math/backends/backend_engine.ts
```diff
@@ -18,61 +18,40 @@
 import * as util from '../../util';
 import {DataType, NDArray} from '../ndarray';
 
-import {MathBackend} from './backend';
-import * as kernel_registry from './kernel_registry';
-import {KernelConfigRegistry} from './kernel_registry';
-import * as tape_util from './tape_util';
-import {ScopeResult, ScopeResultImmediate} from './tape_util';
+export type ScopeResult =
+    void|NDArray|NDArray[]|{[key: string]: NDArray};
```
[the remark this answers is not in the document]
piscisaureus: I agree. I already made it simpler by not having async ScopeResults (a bad idea that wasn't thought through well by the DL authors).

dl/math/math.ts
piscisaureus: Other than removing safeMode (which we discussed), no concerns.

PR #226, opened 2018-02-05
B/tfu2

piscisaureus:


.travis.yml
```diff
@@ -8,6 +8,11 @@ branches:
 language: node_js
 os:
 - linux
+group: edge
```
ry: Why "group: edge" ?

.travis.yml
```diff
@@ -8,6 +8,11 @@ branches:
...
+dist: xenial
+sudo: required
```
ry: Maybe add a comment about why we're using xenial and sudo:required ?

PR #229, opened 2018-02-05
tf binding: add `tensorflowVersion` property

piscisaureus:


piscisaureus:
On Windows (after the tensorflow v1.5.0 upgrade) this says 1.4.0

PR #230, opened 2018-02-05
Use custom tensorflow build

piscisaureus:


ry:
We have no users, so this isn't so important, but I think you've built it with different config than the distributed binaries. In particular you don't have `--copt="-mavx"`
Maybe in the future we can leverage all the docker files TF has: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/ci_build/ci_build.sh

PR #240, opened 2018-02-08
Organize examples

Qti3e:
Hi :)
I moved examples to a separated directory so we can write tons of different examples for the community.
Also, I wrote a simple example in this commit and I'm not sure about the last line (graphing via imshow)
Regards


ry:
Generally a good idea - however we're just about to move all the files around - and nn_example in particular a special case. It's not a real example, but rather a NN library I'm in the middle of developing. I'll try to land the file re-org today.

PR #244, opened 2018-02-09
Fix formatting with negative values.

ry:


format_test.ts
```diff
@@ -31,12 +36,22 @@ test(async function format_2DInt32() {
 });
 
 test(async function format_1DFloat32() {
-  const d = [ 1.2,  0. ,  0. ,  0.123];
-  const actual = toString([4], new Float32Array(d));
-  const expected = "[ 1.2  ,  0.   ,  0.   ,  0.123]";
+  let d = [ 1.2,  0. ,  0. ,  0.123];
+  let actual = toString([4], new Float32Array(d));
```
piscisaureus: const

PR #247, opened 2018-02-09
Reorganize tree

ry:
All core library files are moved to src/, including the binding.

All website files are moved to website/.


Qti3e:
I think you forgot to change [package.json](https://github.com/propelml/propel/blob/r/reorg/package.json#L6)'s main property

ry:
@Qti3e Thanks, Fixed.

PR #248, opened 2018-02-09
Implement focusNextCell

Qti3e:
I closed some of my PRs because I wasn't good enough with git, now I learned git and I start to sending PRs :)
(Sorry for this)

### DIFF
--- a/notebook.ts
+++ b/notebook.ts
@@ -31,6 +31,7 @@ import { assert, delay, IS_WEB } from "./util";
 
 const cellTable = new Map<number, Cell>(); // Maps id to Cell.
 let nextCellId = 1;
+let focusedCellId = null;
 let lastExecutedCellId = null;
 // If you use the eval function indirectly, by invoking it via a reference
 // other than eval, as of ECMAScript 5 it works in the global scope rather than
@@ -141,11 +142,11 @@ export class Cell extends Component<CellProps, CellState> {
       require("codemirror/mode/javascript/javascript.js");
       this.input.innerHTML = "" ; // Delete the <pre>
       this.editor = CodeMirror(this.input, options);
-      this.editor.on("focus", this.focus.bind(this));
-      this.editor.on("blur", this.blur.bind(this));
+      this.editor.on("focus", this.focus.bind(this, false));
+      this.editor.on("blur", this.blur.bind(this, false));
       this.editor.setOption("extraKeys", {
         "Ctrl-Enter": () =>  { this.run(); return true; },
-        "Shift-Enter": () => { this.run(); this.nextCell(); return true; }
+        "Shift-Enter": () => { this.run(); this.focusNextCell(); return true; }
       });
     }
   }
@@ -184,8 +185,23 @@ export class Cell extends Component<CellProps, CellState> {
     if (this.props.onRun) this.props.onRun(this.code);
   }
 
-  nextCell() {
-    // TODO
+  focusNextCell() {
+    const currentId = "cell" + this.id;
+    const cellElements = document.querySelectorAll(".notebook-cell");
+    let found, i = 0;
+    const len = cellElements.length - 1;
+    while (!found && i < len) {
+      if (cellElements[i++].id === currentId) {
+        found = i;
+      }
+    }
+    if (!found) {
+      return;
+    }
+    const cellElement = cellElements[found];
+    lookupCell(Number(cellElement.id.replace("cell", ""))).focus();
+    // TODO: Scroll with a nice animation
+    cellElement.scrollIntoView();
   }
 
   clickedDelete() {
@@ -198,12 +214,25 @@ export class Cell extends Component<CellProps, CellState> {
     if (this.props.onInsertCell) this.props.onInsertCell();
   }
 
-  blur() {
+  blur(shouldBlurEditor = true) {
+    if (focusedCellId === this.id) {
+      focusedCellId = null;
+    }
     this.input.classList.remove("focus");
+    if (shouldBlurEditor) {
+      this.editor.getInputField().blur();
+    }
   }
 
-  focus() {
+  focus(shouldFocusOnEditor = true) {
+    if (focusedCellId) {
+      lookupCell(focusedCellId).blur();
+    }
+    focusedCellId = this.id;
     this.input.classList.add("focus");
+    if (shouldFocusOnEditor) {
+      this.editor.focus();
+    }
   }
 
   render() {
--- a/notebook_test.ts
+++ b/notebook_test.ts
@@ -42,3 +42,32 @@ testBrowser(async function notebook_Notebook() {
   const clones = document.getElementsByClassName("clone");
   assert(0 === clones.length);
 });
+
+testBrowser(async function cell_focusNextCell() {
+  const mdb = enableMock();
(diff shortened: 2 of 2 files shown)


notebook.ts
```diff
@@ -184,8 +184,24 @@ export class Cell extends Component<CellProps, CellState> {
...
+      if (noteBooks[i].id === currentId) {
+        found = true;
+        break;
+      }
+    }
+    if (!found) {
+      return;
+    }
+    const cell = noteBooks[i + 1];
+    lookupCell(Number(cell.id.replace("cell", ""))).focus();
```
ry: When I tried this, it doesn't seem to focus the code mirror. Can you add `this.editor.focus()` in the focus method below?

ry:
LGTM. Please squash your commits.

Qti3e:
Sorry I'm not sure how to write it's testing because I can't access ref of a cell which is inside of Notebook component
```js
testBrowser(async function cell_focusNextCell() {
  const mdb = enableMock();
  document.body.innerHTML = "";

  const readyPromise = new Promise((resolve) => {
    const el = h(nb.Notebook, {
      nbId: "default",
      onReady: resolve,
      ref: (nb) => {
        // make sure we have at least two cells
        nb.onInsertCell(1);
        // How to call lastCell.focusNextCell() ?
        // ...
        resolve();
      }
    });

    render(el, document.body);
  });

  await readyPromise;

  const selectedCell = document.querySelector('#cell1 > .focus')
  assert(1 === selectedCell.length);
});
```

ry:
Try this:

```
  // Test focusNextCell transitions.
  const cells = document.getElementsByClassName("notebook-cell");
  assert(0 < cells.length);
  assert(cells[0].id === "cell1");
  assert(cells[1].id === "cell2");
  const first = nb.lookupCell(1);
  const second = nb.lookupCell(2);
  first.focus();
  assert(first.input.classList.contains('focus'));
  assert(first.editor.hasFocus());
  assert(!second.input.classList.contains('focus'));
  assert(!second.editor.hasFocus());
  first.focusNextCell();
  //FIXME assert(!first.input.classList.contains('focus'));
  assert(!first.editor.hasFocus());
  assert(second.input.classList.contains('focus'));
  assert(second.editor.hasFocus());

```

notebook_test.ts
```diff
@@ -42,3 +42,42 @@ testBrowser(async function notebook_Notebook() {
...
+  assert(cells[1].id === "cell2");
+  const first = nb.lookupCell(1);
+  const second = nb.lookupCell(2);
+  first.focus();
+  assert(first.input.classList.contains('focus'));
+  assert(first.editor.hasFocus());
+  assert(!second.input.classList.contains('focus'));
+  assert(!second.editor.hasFocus());
+  first.focusNextCell();
+  //FIXME assert(!first.input.classList.contains('focus'));
```
ry: Please uncomment and fix this

notebook.ts
```diff
@@ -204,6 +220,7 @@ export class Cell extends Component<CellProps, CellState> {
 
   focus() {
     this.input.classList.add("focus");
+    this.editor.focus();
```
Qti3e: How about this?
```js
focus() {
  cellTable.forEach((cell) => {
    cell.input.classList.remove("focus");
  });
  this.input.classList.add("focus");
  this.editor.focus();
}
```

notebook_test.ts
Qti3e: done

notebook.ts
ry: The loop seems unnecessary. If you add a `this.blur()` to `focusNextCell`, I think it will work.

Qti3e: It's not only about `focusNextCell`, we might change the focused cell later on some other place

notebook.ts
Qti3e: don't you think we should add `editor.getInputField().blur()` to blur method?

ry: Yes, probably. (Is there not just a `editor.blur()` ?)

Qti3e: No, (see)[https://groups.google.com/forum/#!topic/codemirror/C8iQ9bJF3VQ]

notebook_test.ts
```diff
@@ -42,3 +42,41 @@ testBrowser(async function notebook_Notebook() {
...
+testBrowser(async function cell_focusNextCell() {
+  document.body.innerHTML = "";
+
+  const readyPromise = new Promise((resolve) => {
+    const el = h(nb.Notebook, {
+      nbId: "default",
+      onReady: resolve,
+      ref: (nb) => {
+        // make sure we have at least two cells
+        nb.onInsertCell(1);
```
Qti3e: It seems like `onInsertCell` is not working because of Notebook's initial state.

notebook.ts
Qti3e: We can optimize this by always storing the focus element :thinking:

notebook_test.ts
ry: It's unnecessary to grab a ref here.

notebook.ts
```diff
@@ -184,8 +184,24 @@ export class Cell extends Component<CellProps, CellState> {
...
+    for (i = 0; i < len; i++) {
+      if (noteBooks[i].id === currentId) {
+        found = true;
+        break;
+      }
+    }
+    if (!found) {
+      return;
+    }
+    const cell = noteBooks[i + 1];
```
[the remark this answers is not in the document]
Qti3e: What? 😕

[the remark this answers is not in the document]
ry: Sorry - rename `cell` to `cellElement`. (Because `cell` is not a `Cell` but rather an `HTMLElement`)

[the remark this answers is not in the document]

notebook_test.ts
ry: When you query "default", the mock database returns a notebook with a predefined set of cells: 

https://github.com/propelml/propel/blob/b9ee6b1d5ed23e25e5b3f1e42322855eb3f44e64/db.ts#L242-L279

notebook.ts
```diff
@@ -200,10 +216,15 @@ export class Cell extends Component<CellProps, CellState> {
 
   blur() {
     this.input.classList.remove("focus");
+    this.editor.getInputField().blur();
   }
 
   focus() {
+    cellTable.forEach((cell) => {
+      cell.input.classList.remove("focus");
+    });
```
ry: Remove

Qti3e:
Hey Ryan!
I have some bad news
The test fails is not because of this code and we should search for reason somewhere else
Just inject this code into test function:
```js
  ....
  assert(0 < cells.length);
  let a = []
  for (let i = 0;i < cells.length;i++) {
    a.push(cells[i].id)
  }
  console.log(JSON.stringify(a))
  ...
```
Result:
```
 ["cell5","cell6","cell7","cell8"]
```

I can guess it's because `nextCellId` is a global variable and we run multiple test cases without reassigning to it (we should reset this var somewhere).

notebook.ts
Qti3e: will do in next commit

Qti3e:
Closed due to 17652bf819a0b0c2fd2ef95bcda05894d6f702eb

PR #250, opened 2018-02-09
Prevent unintentional cloning

Qti3e:
Imagine this scenario:
A user has a bad internet connection he/she clicks on the clone btn and expects to see sth, but nothing would happen and he/she clicks multiple times.
When he/she connects to the net all requests go one-by-one and that's what actually happened to me when using Notebook for the first time.

### DIFF
--- a/website/notebook.ts
+++ b/website/notebook.ts
@@ -506,6 +506,7 @@ export interface NotebookProps {
 export interface NotebookState {
   doc?: db.NotebookDoc;
   errorMsg?: string;
+  isCloningInProgress: boolean;
 }
 
 // This defines the Notebook cells component.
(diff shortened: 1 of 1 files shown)


notebook.ts
```diff
@@ -525,8 +527,17 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
...
+      return;
+    }
     console.log("Click clone");
+    this.setState({
+      isCloningInProgress: true
+    });
     const clonedId = await db.active.clone(this.state.doc);
+    this.setState({
+      isCloningInProgress: false
+    });
```
ry: nit: can you make these one line

```javascript
 this.setState({ isCloningInProgress: false })
```

PR #256, opened 2018-02-11
[wip] b/nddata

piscisaureus:


piscisaureus:
This patch works but it didn't provide any speedups as I would have hoped it would. Closing.

PR #270, opened 2018-02-17
API improvements (rescale, params.init, params.scope, sgd)

ry:


src/tensor.ts
```diff
@@ -550,6 +552,69 @@ export class Tensor implements types.BasicTensor {
...
+  /** To rescale a tensor from inScale to outScale.
+   * This is often used rescale images from [0, 255] integer
+   * range to [-1, 1] range. This function will automatically cast tensors to
+   * float32.
+   *
+   *    import { T } from "propel";
+   *    let pretendImage = T([0, 127, 255], "int32");
+   *    pretendImage.rescale([0, 255], [-1, 1]);
+   */
+  rescale(inScale: number[], outScale: number[]): Tensor {
```
piscisaureus: Use type `[number, number]`

src/tensor.ts
```diff
@@ -550,6 +552,69 @@ export class Tensor implements types.BasicTensor {
...
+   * If users are calling this function multiple times, they should use
+   * params.scope("myLayerName") to not conflict with other params.
+   *
+   *    import { params, zeros } from "propel";
+   *    params = params();
+   *    inputs = zeros([2, 5]);
+   *    outputs = inputs.linear(params.scope("L1"), 10);
+   *    params.has("L1/weights") && params.has("L1/bias");
+   */
+  linear(params: Params, outDim: number, opts: LinearOpts = {}): Tensor {
```
piscisaureus: Note that you can place a default value for all opts here:

```ts
linear(params: Params, outDim: number, { bias = true, scale = 0.01 }: LinearOpts = {}): Tensor {
```

src/api.ts
```diff
@@ -157,84 +163,32 @@ export function ones(shape: types.Shape,
...
-    // Initialize.
-    let t = zeros(shape);
-    if (device && device !== "CPU:0") {
-      t = new Tensor(bo.copyToDevice(t.basic, device));
+    for (const name of Object.keys(grads)) {
+      const g = grads[name];
+      const p = params.get(name);
+      // p -= g * lr
+      p.assign(p.sub(g.mul(opts.lr)));
+      keep(p);
```
piscisaureus: Is keep() necessary here. I would expect the tensor to be kept implicitly, since it existed before the call to gc().

ry: Good point - removed.

src/tensor.ts
ry: done

src/tensor.ts
ry: done

PR #271, opened 2018-02-17
Fix bug where insert cell button deletes other output.

ry:
[Currently failing tests. Will fix.]


ry:
closed in favor of #360

PR #273, opened 2018-02-18
Rename T() to tensor()

piscisaureus:


piscisaureus:
After landing this, check that the example notebooks still work.

piscisaureus:
TODOs after landing this:

* update the example notebooks
* re-run gendoc
* release new npm package

piscisaureus:
Note that the tests are failing ("Params is not a constructor"), which is unrelated to this PR.
This should be fixed by re-running gendoc.
I'm not sure why the CI didn't catch this when testing #270.

src/api.ts
```diff
@@ -163,7 +167,7 @@ export function ones(shape: types.Shape,
   return ops.ones(shape, opts);
 }
 
-interface SGDOpts {
+export interface SGDOpts {
```
ry: Unrelated change?

website/docs.ts
```diff
@@ -125,9 +125,6 @@ function startsWithUpperCase(s: string): boolean {
 export function Docs(props) {
   let docs: DocEntry[] = require("./docs.json");
   docs = docs.sort((a, b) => {
-    // Special case "T" to be at the top of the docs.
-    if (a.name === "T") return -1;
-    if (b.name === "T") return 1;
```
ry: can we leave it at the top of the docs?

piscisaureus: Of course we can, but I thought it looked kinda weird.

ry: ok nevermind

PR #274, opened 2018-02-19
tf binding: tell v8 about the size of tensor handles

piscisaureus:
This deals with the ridiculous memory spikes and long GC pauses when
using the tensorflow backend.

Note that there still seem to be memory leaks when running nn_example_main, but only slowly. 
A simple test program doesn't seem to leak:

```
const t1 = api_1.T(new Float32Array(1e6), [1000, 1000]);
const t2 = api_1.T(new Float32Array(1e6), [1000, 1000]);
for (;;) {
  const t3 = t1.mul(t2);
}
```

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -76,6 +76,27 @@ class JSRef {
   napi_ref ref_;
 };
 
+static int64_t GetHandleByteSize(TFE_TensorHandle* h) {
+  int64_t size = TF_DataTypeSize(TFE_TensorHandleDataType(h));
+  int num_dims = TFE_TensorHandleNumDims(h);
+  for (int i = 0; i < num_dims; i++) {
+    size *= TFE_TensorHandleDim(h, i);
+  }
+  return size;
+}
+
+static void RegisterHandle(napi_env env, TFE_TensorHandle* h) {
+  int64_t size = GetHandleByteSize(h);
+  int64_t total;
+  napi_adjust_external_memory(env, size, &total);
+}
+
+static void UnregisterHandle(napi_env env, TFE_TensorHandle* h) {
+  int64_t size = GetHandleByteSize(h);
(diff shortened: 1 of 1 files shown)


src/binding.cc
```diff
@@ -76,6 +76,27 @@ class JSRef {
   napi_ref ref_;
 };
 
+static int64_t TFTensorHandleSize(TFE_TensorHandle* h) {
```
ry: On the naming:
1. `s/Size/ByteSize`
2. It's a bit confusing to start non-TF functions with `TF`.

I'd suggest renaming to `HandleByteSize`

src/binding.cc
```diff
@@ -76,6 +76,27 @@ class JSRef {
...
+  int64_t size = TF_DataTypeSize(TFE_TensorHandleDataType(h));
+  int num_dims = TFE_TensorHandleNumDims(h);
+  for (int i = 0; i < num_dims; i++) {
+    size *= TFE_TensorHandleDim(h, i);
+  }
+  return size;
+}
+
+static void TFTensorHandleRegister(napi_env env, TFE_TensorHandle* h) {
```
ry: suggest the function gets renamed with `s/TFTensor//`

src/binding.cc
piscisaureus: done

src/binding.cc
piscisaureus: done

piscisaureus:
PTAL

src/binding.cc
ry: I don’t see any changes?

piscisaureus: Sorry, got some branches mixed up.

piscisaureus: Pushed it again.

PR #278, opened 2018-02-20
New dataset interface based on tf.data.Dataset

ry:


src/dataset.ts
```diff
@@ -0,0 +1,203 @@
...
+    const sliceSize = Math.min(batchSize, this.batchDim - this.pos);
+
+    for (const name of Object.keys(this.tensors)) {
+      const t = this.tensors[name];
+      // TODO This is ugly. slice() should take truncated begin and size
+      // arguments.
+      const begin: number[] = [];
+      const size: number[] = [];
+      for (let i = 0; i < t.rank; i++) {
+        if (i === 0) {
```
piscisaureus: This works of course, but it's such an overly complicatged way of doing it.
Why not:
```js
const begin =[this.pos, ...Array(t.rank - 1).fill(0)];
const size = [sliceSize, ...Array(t. rank - 1).fill(-1)];
```

src/dataset.ts
```diff
@@ -0,0 +1,203 @@
...
+
+class LoadingSlicesDataset extends SliceDataset {
+  constructor(private promise: Promise<NamedGroupOf<Tensor>>) {
+    super(null);
+    promise.then((tensors) => {
+      this.tensors = tensors;
+      this.checkTensors();
+      this.promise = null;
+    }).catch((e) => {
+      throw e;
```
piscisaureus: This does nothing.

src/dataset.ts
```diff
@@ -0,0 +1,203 @@
...
+        }
+      }
+      out[name] = t.slice(begin, size);
+    }
+    this.pos += batchSize;
+    return out;
+  }
+}
+
+class LoadingSlicesDataset extends SliceDataset {
```
piscisaureus: * Why is this a separate class? Can't `class SliceDataSet` have a constructor that takes an argument with type `NamedGroupOf<Tensor>|Promise<NamedGroupOf<Tensor>>` ?
* It's odd that the parent class is SliceDataSet but this class is named ...Slice**s**DataSet

src/dataset.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+export function dataset(name: string, split = "train"): Dataset {
+  if (name !== "mnist") throw Error("Bad dataset " + name);
+
+  return new LoadingSlicesDataset(mnist.loadSplit(split));
+}
+
+abstract class Dataset {
+  constructor(protected parent?: Dataset) { }
```
piscisaureus: What's `parent` for? I only see it used in the `reset` call.

src/dataset.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  abstract async next(): Promise<NamedGroupOf<Tensor>>;
```
piscisaureus: Out of scope for this patch - but I think it'd be nice to implement [Symbol.iterator()] or [Symbol.asyncIterator()] in addition to having a bespoke next() function.

src/dataset.ts
ry: datasets are meant to be chained

```
dataset("mnist").repeat().batch(128).shuffle()
```

parent points to the parent.

src/dataset.ts
ry: Combined LoadingSlicesDataset functionality into SliceDataset.

src/dataset.ts
```diff
@@ -0,0 +1,203 @@
...
+   limitations under the License.
+ */
+
+// This module is inspired by TensorFlow's tf.data.Dataset.
+// https://www.tensorflow.org/api_docs/python/tf/data/Dataset
+import { isUndefined } from "util";
+import * as mnist from "./mnist";
+import { Tensor } from "./tensor";
+
+type NamedGroupOf<T> = { [name: string]: T };
```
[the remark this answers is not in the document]
ry: Left over from an earlier version where it was needed. Renamed to `NamedTensors`.

src/dataset.ts
ry: Removed.

src/dataset.ts
ry: Fixed.

src/dataset.ts
ry: Agree.

src/dataset.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  async next(): Promise<NamedTensors> {
+    while (true) {
+      const r = await this.parent.next();
+      if (r != null) {
+        return r;
+      } else {
+        this.epoch++;
+        if (this.count && this.epoch >= this.count) {
```
piscisaureus: Kind of nitty but I think that if count is 0 (as opposed to null or undefined), then next should repeat null immediately.

ry: Throwing on 0 is easier. PTAL.

PR #279, opened 2018-02-20
Calls gendoc during build_website.

ry:


ry:
Changes included in #280.

PR #280, opened 2018-02-20
Fix params_test. Run gendoc during build_website.

ry:
Bug introduced in https://github.com/propelml/propel/commit/6870a617a431ab3dfc6348d1eefd811e2b08da88

The test wasn't included in test_isomorphic.


src/api.ts
```diff
@@ -228,7 +228,7 @@ export class OptimizerSGD {
       for (const name of Object.keys(grads)) {
         const g = grads[name];
         const p = this.params.get(name);
-        const v = this.velocity.init(name, () => zeros(p.shape));
+        const v = this.velocity.define(name, () => zeros(p.shape));
```
piscisaureus: Nice!
I actually wanted to suggest this when I was reviewing the params patch, but didn't think it was important enough.
Liking the validation that we independently came up with the same idea.

.gitignore
```diff
@@ -26,18 +26,11 @@
 /build/
 /CMakeFiles/
 /CMakeScripts/
-/Debug/
```
piscisaureus: Why did you remove this, but not CMake*, MiniSizeRel, RelWithDebInfo, Win32, x64 etc?

website/website_main.ts
```diff
@@ -8,7 +8,9 @@ assert(IS_WEB);
 
 enableFirebase();
 
-window.addEventListener("load", () => {
+window.addEventListener("load", async() => {
   render(h(Router, null), document.body, document.body.children[0]);
-  drainExecuteQueue();
+  await drainExecuteQueue();
+  rerender();
```
piscisaureus: * Can you make sure this only happens on CI? Use `if (window.navigator.webdriver) {`
* Are you sure this actually happens after everything else is done? I could imagine that if there's a setTimeout() or fetch() during cell evaluation, it might trigger another render *after* this rerender() call.

ry: Added the branch. No, a setTimeout would still break this, we're just doing a best effort to run the cells.

.gitignore
ry: Added a new commit cleaning up gitignore.

PR #283, opened 2018-02-21
dl: optimize cpu matmul

piscisaureus:
On my machine, on the order of 100x speedup.
To try: `PROPEL=dl ts-node matmul_bench.ts`.
I'm not sure where the benchmark should go.

### DIFF
--- a/src/dl/math/backends/backend_cpu.ts
+++ b/src/dl/math/backends/backend_cpu.ts
@@ -337,30 +337,39 @@ export class MathBackendCPU implements MathBackend {
     const rightDim =
         (bOrientation === MatrixOrientation.REGULAR) ? b.shape[1] : b.shape[0];
 
-    const normalGetter = (matrix: Array2D, i: number, j: number) =>
-        matrix.get(i, j);
-    const transposedGetter = (matrix: Array2D, i: number, j: number) =>
-        matrix.get(j, i);
+    const aValues = a.dataSync();
+    const bValues = b.dataSync();
 
-    const aGetter = (aOrientation === MatrixOrientation.REGULAR) ?
-        normalGetter :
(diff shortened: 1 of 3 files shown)


src/dl/math/backends/backend_cpu.ts
```diff
@@ -337,30 +337,37 @@ export class MathBackendCPU implements MathBackend {
...
-        transposedGetter;
-    const bGetter = (bOrientation === MatrixOrientation.REGULAR) ?
-        normalGetter :
-        transposedGetter;
-    const values = new Float32Array(leftDim * rightDim);
-    let index = 0;
-    for (let i = 0; i < leftDim; ++i) {
-      for (let j = 0; j < rightDim; ++j) {
+    const aValues = a.dataSync();
+    const bValues = b.dataSync();
```
ry: I’m concerned about the sync here. Does this patch speed up testNNExample?

piscisaureus: Yes, a lot.
Also this code is cpu only - dataSync is basically a no-op.

ry: oh right. can confirm 100x speed up on mac.

matmul_bench.ts
```diff
@@ -0,0 +1,22 @@
+
+import { tensor } from "./src/api";
+
+const t1 = tensor(new Array(100).fill(new Array(100).fill(1)));
+const t2 = tensor(new Array(100).fill(new Array(100).fill(2)));
```
ry:
```
const t1 = ones([100, 100]);
const t2 = ones([100, 100]).add(1);
```

piscisaureus: fixed

PR #285, opened 2018-02-22
Make linear take scopeName argument.

ry:
This forces scoping to happen, while still being explicit in parameter naming. It's also less characters than params.scope("...").


ry:
closed in favor of #286

PR #286, opened 2018-02-22
API improvements

ry:
This forces scoping to happen, while still being explicit in parameter naming. It's also less characters than params.scope("...").


piscisaureus:
The "done" situation is pretty ugly, but I can see how there isn't really a better solution to this.

PR #287, opened 2018-02-23
Adds new example NN API. Removes old.

### LINKED ISSUE #259: NN API
We've been putting off building up the NN API until we had paid down some technical debt. But it's time to expand again. I want a small example for the home page, so it must be less than 10 lines.

Here's a proposal for training an mnist classifier which saves checkpoints

https://gist.github.com/ry/9d91254c1d03e416ce4b751b711b4b71

ry:
fixes #259


example.js
```diff
@@ -0,0 +1,56 @@
...
+        // Apply three linear (densely connected layers) with a relu after each
+        // except the last. The shape of the activations:
+        // [128, 28, 28] -> [128, 200] -> [128, 100] -> [128, 10] -> []
+        // Note the final loss is a scalar.
+        .linear("L1", params, 200).relu()
+        .linear("L2", params, 100).relu()
+        .linear("L3", params, 10)
+        // Using the logits, calculate a classification loss between the
+        // labels. Labels's shape is [128]. This value is returned, and will be
+        // backpropagated though.
```
piscisaureus: through.

src/experiment.ts
```diff
@@ -0,0 +1,144 @@
...
+ */
+export async function print(...args: PrintArgs): Promise<void> {
+  // Drop prints if we get too backed up.
+  if (printWaitCount > 100) {
+    console.log("drop print");
+    await delay(0);
+    return;
+  }
+  const promise = printHelper(...args);
+  // All this is to make sure that log items are in the order they came.
```
piscisaureus: This is very complicated. I would suggest the following:

* lastPrint is *always* a promise. You can make an empty resolved promise with `Promise.resolve()`.
* Unconditionally do the following: `lastPrint = lastPrint.then(() => printHelper(...args))`

src/experiment.ts
```diff
@@ -0,0 +1,144 @@
...
+  });
+  await promise;
+}
+
+async function printHelper(...args: PrintArgs): Promise<string> {
+  const strings = await Promise.all(args.map(async(arg) => {
+    if (typeof arg === "string") {
+      return arg;
+    } else if (typeof arg === "number") {
+      return String(arg);
```
piscisaureus: This conversion is unnecessary, and this if...elseif...else block is missing cases for certain cases like booleans, null etc.
I would suggest to check for tensors explicitly, and otherwise just return the input.
```
if (arg instanceof Tensor) {
  return format.toString(arg.shape, await arg.data());
} else {
  return arg;
}
```

src/experiment.ts
```diff
@@ -0,0 +1,144 @@
...
+ * them. The following code will not block the browser.
+ *
+ *    import * as pr from "propel"
+ *    pr.print("hello", pr.range(3));
+ */
+export async function print(...args: PrintArgs): Promise<void> {
+  // Drop prints if we get too backed up.
+  if (printWaitCount > 100) {
+    console.log("drop print");
+    await delay(0);
```
piscisaureus: Why this delay?

src/experiment.ts
```diff
@@ -0,0 +1,144 @@
...
+      // Forward/Backward pass
+      const gradsAndLoss = gradFn(this.currentParams);
+      const grads = gradsAndLoss[0];
+      loss = gradsAndLoss[1];
+      assert(loss.rank === 0);
+      keep(loss);
+      optimizer(opts, this.currentParams, grads);
+    });
+
+    await print("step", this.step, "loss", loss);
```
piscisaureus: The `await` statement here defeats the purpose of throttling in your `print` function.
So maybe it may not make a lot of sense to address my comments above, because why do it at all?

ry: Made async.

src/experiment.ts
ry: Cruft from earlier debugging. Removed.

src/experiment.ts
ry: done

src/experiment.ts
ry: Done.

ry:
by snapshotting do you mean checkpointing? that will be done every N steps or every so many minutes - not every step.

I think the printing backoff isn't working because TF's tensor.data() is actually synchronous, same for DL CPU. It probably only happens when using webgl. However, given that we can't test it, I've removed that code. 

I've also updated the maxSteps to 3k.

Another important point is that this network hasn't been tested on GPU yet - while nn_example has.. Ideally this example would default to using GPUs without additional code.

Issue #196, opened 2018-01-30
Support RGB in imshow


ry:

Qti3e:
I would like to work on this one : )

ry:
You're welcome to it - thanks Parsa!

qti3e:
Deps on #311
https://gist.github.com/qti3e/f516ad5c597e811971d0443b3261a151

qti3e:
It should be closed : )

PR #291, opened 2018-02-27
[WIP] Implement imread

Qti3e:


src/imparser.ts
```diff
@@ -0,0 +1,136 @@
...
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+
+// This file is to parse image files,
+// It uses Canvas on the web and on Node.js it'll use `pngjs` and `jpeg-js`
```
ry: I think `imread.ts` or `image.ts` would be a better name for this file.

src/imparser.ts
```diff
@@ -0,0 +1,136 @@
...
+      };
+      img.src = src;
+    } catch (e) {
+      resolve();
+    }
+  });
+}
+// </IS_WEB>
+
+// <IS_NODE>
```
ry: Use `if (IS_NODE)`

src/imparser.ts
```diff
@@ -0,0 +1,136 @@
...
+const imageSigns: Array<[number[], string]> = [
```
ry: is "signs" the right terminology? I think "header" or "prefix" would be more appropriate.

src/imparser.ts
```diff
@@ -0,0 +1,136 @@
...
+
+function readMIME(src: string): string {
+  const fs = nodeRequire("fs");
+  // read file header
+  const header = new Buffer(signMaxLength);
+  const fd = fs.openSync(src, "r");
+  fs.readSync(fd, header, 0, signMaxLength);
+  fs.closeSync(fd);
+  // find matched MIME type
+  for (let i = 0; i < imageSigns.length; i += 1) {
```
ry: nit: `++i` instead of `i+=1`. same for the other loops.

src/imparser.ts
```diff
@@ -0,0 +1,136 @@
...
+// </IS_WEB>
+
+// <IS_NODE>
+const imageSigns: Array<[number[], string]> = [
+  [[0xFF, 0xD8], "image/jpeg"],
+  [[0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A], "image/png"]
+];
+const signMaxLength: number = Math.max(...imageSigns.map(x => x[0].length));
+
+function readMIME(src: string): string {
```
ry: rename `src` to `filename`

src/imparser.ts
```diff
@@ -0,0 +1,136 @@
...
+  }
+  return null;
+}
+
+// <FileHandlers>
+function PNGHandler(src: string): Promise<types.RawImage> {
+  const PNG = nodeRequire("pngjs").PNG;
+  const fs = nodeRequire("fs");
+  return new Promise(resolve => {
+    fs.createReadStream(src)
```
ry: No need to be async in filesystem stuff in Node. Just use `fs.readFileSync`. It's usually faster.

src/imparser_test.ts
```diff
@@ -0,0 +1,45 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+
+import { test } from "../tools/tester";
+import { imageDecoder } from "./imparser";
+import * as types from "./types";
+import { assert } from "./util";
+
+// test is valid with this files:
```
ry: nit: use proper capitalization, and I would rephrase: `The tests use these files:`

src/imparser_test.ts
```diff
@@ -0,0 +1,45 @@
...
+// fa57d083e48e999ed3f210aefd92e5f7  testdata/sample.png
+// d37ff170a58223de46152ce837b8e0c4  testdata/sample.jpg
+
+test(async function readPNG() {
+  const img: types.RawImage = await imageDecoder("./testdata/sample.png");
```
ry: Use `path.join(__dirname, "testdata/sample.png")`

src/types.ts
```diff
@@ -111,3 +111,11 @@ export interface TensorOpts {
   dtype: DType;
   device?: string;
 }
+
+// Raw image format which is read from disck or canvas
```
ry: disk. And use proper punctuation.

src/types.ts
```diff
@@ -111,3 +111,11 @@ export interface TensorOpts {
...
+export interface RawImage {
+  width: number;
+  height: number;
+  // TypedArray<r, g, b, a, r, g, b, a, ...>
+  data: TypedArray | Buffer;
```
ry: Buffer is node only, so don't use that here. Also, I don't think there's any benefit in adding this `RawImage` interface over having a rank three tensor. For RGBA, it would have the shape `[4, height, width]` and have dtype `int32` (we don't yet support `uint8`).

src/imparser.ts
```diff
@@ -0,0 +1,136 @@
...
+// </IS_NODE>
+
+export async function imageDecoder(src: string): Promise<types.RawImage> {
+  if (IS_WEB) {
+    return await WebImageDecoder(src);
+  }
+  return await NodeImageDecoder(src);
+}
+
+export async function imageEncoder(image: types.RawImage)
```
ry: Leave this out until it's implemented.

src/imparser.ts
```diff
@@ -0,0 +1,136 @@
...
+
+function toExactRawImage(img: types.RawImage): types.RawImage {
+  return {
+    width: img.width,
+    height: img.height,
+    data: img.data
+  };
+}
+
+async function NodeImageDecoder(src: string): Promise<types.RawImage> {
```
ry: Return a `Promise<Tensor>` here.

src/imparser.ts
```diff
@@ -0,0 +1,136 @@
...
+  return new Promise(resolve => {
+    fs.createReadStream(src)
+      .pipe(new PNG())
+      .on("parsed", function(this: types.RawImage) {
+        resolve(this);
+      });
+  });
+}
+
+function JPEGHandler(src: string): Promise<types.RawImage> {
```
ry: Return `Promise<Tensor>` here.

src/imparser.ts
Qti3e: [List of file signatures](https://en.wikipedia.org/wiki/List_of_file_signatures)

src/imparser.ts
Qti3e: Cause TypeScript to produce error
```
Cannot find name 'NodeImageDecoder'. Did you mean 'WebImageDecoder'?
Change spelling to 'WebImageDecoder'
```

src/imparser.ts
ry: `imageSignatures` works.

src/imparser.ts
Qti3e: You mean I should move all logic of [matplotlib.ts:imshow](https://github.com/propelml/propel/pull/291/files#diff-ee33f25298e461a62b32c2a1ad689879R193) into this function?
it would be duplicated in both `NodeImageDecoder` and `JPEGHandler`

Qti3e: That's exactly what I did:
image
But if you want I can also remove the RawImage section:
image

ry: Yes, please.

Qti3e: done

Qti3e:
How about docs for `imread` function?

src/image.ts
```diff
@@ -0,0 +1,135 @@
...
+
+// This file is to parse image files,
+// It uses Canvas on the web and on Node.js it'll use `pngjs` and `jpeg-js`
+
+import { Tensor } from "./api";
+import { convert } from "./tensor";
+import { IS_NODE, nodeRequire } from "./util";
+import {  emptyArray } from "./util";
+
+function Int32ToTensor(data: Int32Array, height: number, width: number)
```
ry: Doing the loop below is slow - on TensorFlow it can be done in the GPU. The `convert()` can make a TypedArray directly into a Tensor, then you can use `transpose` to change the axes around. Try something like this:
```javascript
convert(data).reshape([height, width, 4]).transpose([2, 0, 1])
```
(warning: untested code)

src/image.ts
```diff
@@ -0,0 +1,135 @@
...
+      return PNGHandler(filename);
+    case "image/jpeg":
+      return JPEGHandler(filename);
+    default:
+      throw new Error("This file is not a valid PNG/JPEG image.");
+  }
+}
+// </IS_NODE>
+
+export async function imread(filename: string): Promise<Tensor> {
```
ry: It would be nice if the user could specify if they want "RGB" or "RGBA" like is done in python: https://docs.scipy.org/doc/scipy/reference/generated/scipy.misc.imread.html

Also please add some documentation with an example. The documentation always runs in the web browser, so it would be cool to show it loading a remote png:
```
let tensor = await imread("https://avatars2.githubusercontent.com/u/80?s=40&v=4")
imshow(tensor)
```
(We can't use imshow in the docs current, unfortunately,  (#199) so maybe just display the shape of the loaded image tensor)

Qti3e:
Oh... 
After I switched to `PNG.sync.read(data)` instead of old:
```js
  const PNG = nodeRequire("pngjs").PNG;
  const fs = nodeRequire("fs");
  return new Promise(resolve => {
    fs.createReadStream(src)
      .pipe(new PNG())
      .on("parsed", function(this: types.RawImage) {
        resolve(this);
      });
  });
```
It throw a weired error 
```
/usr/bin/node[27466]: ../src/node_zlib.cc:189:static void node::{anonymous}::ZCtx::Write(const v8::FunctionCallbackInfo<v8::Value>&) [with bool async = false]: Assertion `Buffer::HasInstance(args[4])' failed.
 1: node::Abort() [/usr/bin/node]
 2: 0x8c811b [/usr/bin/node]
 3: 0x93f065 [/usr/bin/node]
 4: v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&)) [/usr/bin/node]
 5: 0xaddc5c [/usr/bin/node]
 6: v8::internal::Builtin_HandleApiCall(int, v8::internal::Object**, v8::internal::Isolate*) [/usr/bin/node]
 7: 0x19f9208042fd
[2]    27456 abort (core dumped)  ts-node ./im_test.ts
```

Qti3e:
It's Node v9.x.x compatibility error lukeapage/pngjs#96
What should I do? use async version or skip Node 9...

src/image.ts
Qti3e: I think it's because of #196, not #199

Qti3e: done

src/im_test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+// The tests use these files:
+// fa57d083e48e999ed3f210aefd92e5f7  testdata/sample.png
+const pngPath = path.join(__dirname, "testdata/sample.png");
+// d37ff170a58223de46152ce837b8e0c4  testdata/sample.jpg
+const jpgPath = path.join(__dirname, "testdata/sample.jpg");
+
+test(async function readPNG() {
+  const img = await imread(pngPath);
+  assertShapesEqual([4, 64, 64], img.shape);
```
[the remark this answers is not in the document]
Qti3e: done

src/test_isomorphic.ts
```diff
@@ -2,6 +2,7 @@ import "./api_test";
 import "./backend_test";
 import "./dataset_test";
 import "./format_test";
+import "./im_test";
```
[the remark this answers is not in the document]
Qti3e: done

src/util.ts
```diff
@@ -345,3 +345,17 @@ export async function fetchArrayBuffer(path: string): Promise<ArrayBuffer> {
 export async function fetchStr(path: string): Promise<string> {
   return fetch2(path, "utf8") as any;
 }
+
+export function emptyArray(shape: Shape): RegularArray<number> {
```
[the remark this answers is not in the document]
Qti3e: done

src/image.ts
Qti3e: It works for RGBA, how about RGB?

Qti3e: done by appending `.slice([0, 0, 0], [3, height, width])`

src/image.ts
```diff
@@ -0,0 +1,147 @@
...
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+
+// This file is to parse image files,
+// It uses Canvas on the web and on Node.js it'll use `pngjs` and `jpeg-js`
```
ry: This module allows Propel to read PNG and JPG images.

src/image.ts
```diff
@@ -0,0 +1,147 @@
...
+
+function Int32ToTensor(data: Int32Array, height: number, width: number,
+type: "RGB" | "RGBA"): Tensor {
+  if (type === "RGB") {
+    return convert(data)
+      .reshape([height, width, 4])
+      .transpose([2, 0, 1])
+      .slice([0, 0, 0], [3, height, width]);
+  }
+  return convert(data).reshape([height, width, 4]).transpose([2, 0, 1]);
```
ry: 1. Rename "type" to "mode"
2. The reshape and transpose are unnecessarily duplicated.
3. `slice(0, 3)` should work if you rebase onto master.
4. Rename `Int32ToTensor` to `toTensor`
5. What about grayscale images?

src/image.ts
```diff
@@ -0,0 +1,147 @@
...
+      }
+    }
+    if (ret) {
+      return type;
+    }
+  }
+  return null;
+}
+
+// <FileHandlers>
```
ry: Please remove these html tags comments. (I only added those in the example.ts for extra clarity - it's not standard practice.)

src/image.ts
Qti3e: done

src/image.ts
Qti3e: done

PR #292, opened 2018-02-27
Sandbox notebook execution

piscisaureus:


tools/build_website_impl.ts
```diff
@@ -76,6 +76,7 @@ process.on("unhandledRejection", e => { throw e; });
 
   await writePages();
   await run.parcel("website/website_main.ts", "build/website");
+  await run.parcel("website/nb_sandbox.ts", "build/website");
```
ry: Why does this need to be a separate from website_main bundle?

website/nb_rpc_channel.ts
```diff
@@ -0,0 +1,131 @@
+
```
ry: Copyright header and a little explanation would be nice here.

website/nb_rpc_channel.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+interface ReturnMessage extends Message {
+  type: "return";
+  id: string;
+  result?: any;
+  exception?: any;
+}
+
+interface Resolver<T> extends Promise<T> {
```
ry: I don't understand this.. How is this different than Promise<T>?

website/nb_rpc_channel.ts
```diff
@@ -0,0 +1,131 @@
...
+  resolve: (value?: T) => void;
+  reject: (value: any) => void;
+}
+
+function createResolver<T>(): Resolver<T> {
+  let methods;
+  const promise = new Promise((...args) => { methods = args; }) as Resolver<T>;
+  [promise.resolve, promise.reject] = methods;
+  return promise;
+}
```
ry: I don't understand what this is doing. Can you add a comment?

website/nb_sandbox.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (m) {
+    return m;
+  }
+  throw new Error("Unknown module: " + target);
+}
+
+console.log("load");
+
+const rpc = new RpcChannel(window.parent, {
+  async evalCell(source: string, cellId: number): Promise<any> {
```
ry: I think you can `s/Promise<any>/Promise/`

website/nb_sandbox.ts
```diff
@@ -0,0 +1,63 @@
...
+
+// If you use the eval function indirectly, by invoking it via a reference
+// other than eval, as of ECMAScript 5 it works in the global scope rather than
+// the local scope. This means, for instance, that function declarations create
+// global functions, and that the code being evaluated doesn't have access to
+// local variables within the scope where it's being called.
+const globalEval = eval;
+const global = globalEval("this");
+
+class Console {
```
ry: Nice. This looks much saner than what we had.

piscisaureus:
@Qti3e Yes -- we have to do this now, because the notebook is susceptible to cross-site scripting. E.g. I could create a malicious notebook, and if you were to look at it, it'd delete your notebooks.

tools/build_website_impl.ts
piscisaureus: It's the script that runs in the iframe.
The iframe is hidden and it renders no UI, hence to reason to load preact etc.
However it does need the propel lib (which the main website bundle now no longer needs).

website/nb_rpc_channel.ts
piscisaureus: It's a Promise extended with two methods: `reject()` and `resolve()`.

tools/build_website_impl.ts
ry: Maybe add a comment explaining that here?

website/notebook.ts
```diff
@@ -53,6 +45,48 @@ export function lookupCell(id: string | number) {
...
 
+function createSandbox(): RpcChannel {
+  const iframe = document.createElement("iframe");
+  iframe.setAttribute("sandbox", "allow-scripts");
+  iframe.setAttribute("srcdoc", `<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <script type="text/javascript"
+            src="${window.location.origin}/nb_sandbox.js"></script>
```
ry: When rendering with JSDOM `window.location.origin` is `null`

src/im.ts
```diff
@@ -154,5 +154,10 @@ export async function imread(filename: string, mode: Mode = "RGBA")
   if (IS_NODE) {
     return await nodeImageDecoder(filename, mode);
   }
+  // Neither image decoder works in JSDOM while building the website.
+  // TODO: fix this.
```
ry: I created an issue for this https://github.com/propelml/propel/issues/309 - please link here.

website/nb_sandbox.ts
```diff
@@ -0,0 +1,118 @@
...
+      // When running tests, rethrow any errors. This ensures that errors
+      // occurring during notebook cell evaluation result in test failure.
+      if (window.navigator.webdriver) {
+        throw e;
+      }
+    }
+  }
+});
+
+function guessCellId(): number {
```
ry: This function is duplicated in notebook.ts

src/matplotlib.ts
```diff
@@ -12,145 +12,25 @@
...
+let currentOutputHandler: OutputHandler = null;
+export function setOutputHandler(h: OutputHandler) {
+  if (currentOutputHandler !== null) {
+    console.warn("Warning: replacing existing OutputHandler");
   }
-  h = handler;
+  currentOutputHandler = h;
 }
 
-export function outputEl(): null | Element {
```
ry: outputEl still exists in notebook.ts but is unused.

tools/build_website_impl.ts
```diff
@@ -77,8 +97,13 @@ process.on("unhandledRejection", e => { throw e; });
 
   scss("website/main.scss", join(websiteRoot, "bundle.css"));
 
-  await writePages();
   await run.parcel("website/website_main.ts", "build/website");
+  await run.parcel("website/nb_sandbox.ts", "build/website");
```
ry: Could you add a comment for each of these bundles, describing their purpose?

website/nb_rpc_channel.ts
ry: Don't promises already have those?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject

src/im.ts
Qti3e: Image decoder works in JSDOM if you suppose that as a Node thing, I suggest adding this test condition to line 154:
```js
if (IS_NODE) {
...
}
```

website/nb_rpc_channel.ts
piscisaureus: Promise.resolve creates a promise that's already resolved.
This is something else:

```
const p = createResolvable(); // Creates a promise.
p.resolve(); // Resolve promise p.
```

tools/build_website_impl.ts
piscisaureus: done

website/nb_rpc_channel.ts
piscisaureus: done

website/nb_rpc_channel.ts
piscisaureus: Added a bunch of comments, moved it to util.ts.

website/nb_rpc_channel.ts
piscisaureus: done

website/nb_sandbox.ts
piscisaureus: No longer relevant -- it now return Promise<void>

website/notebook.ts
piscisaureus: fixed

src/im.ts
piscisaureus: It doesn't currently work, because the `nodeImageDecoder` calls `fs.readFile`, which is not available in the JSDOM sandbox. I can be made to work of course - will address this later.

website/nb_sandbox.ts
piscisaureus: Fixed

src/matplotlib.ts
piscisaureus: Fixed - removed the unused function.

tools/build_website_impl.ts
piscisaureus: Done

src/im.ts
qti3e: Oh right, I didn't think about that.

PR #296, opened 2018-02-28
For review: add_ function

ryanmurakami:
FIrst pass, looking for feedback and thoughts on the approach. 

Open question: I'm just returning `x` to `add_`. Is this sufficient or does `x` need to be protected? 

Let me know your thoughts!


src/ops.ts
```diff
@@ -178,6 +178,12 @@ defBW("add",
   (g, sx, sy) => addGrad(true)(g, sx, sy),
   (g, sx, sy) => addGrad(false)(g, sx, sy));
 
+export const add_ = (x: Tensor, y: Tensor): Tensor => {
```
ry: Do this just inside `tensor.ts`. It might make sense to put into ops, but I think if it can only exist at the surface layer, it'd be better.

src/ops_test.ts
```diff
@@ -0,0 +1,30 @@
...
+   limitations under the License.
+ */
+import { test } from "../tools/tester";
+import { tensor } from "./api";
+import * as ops from "./ops";
+import { assertAllEqual } from "./util";
+
+// Basic Tests
+
+test(async function ops_add_() {
```
ry: Move this test to `api_test.ts`(it can go directly after `api_inc()`)

ryanmurakami: Got it. I was thinking if the ops start expanding a lot, it might be cleaner to have a separate test file for them. Anyways, I'll be moving the in-place ops to to `tensor.ts` as suggested, so `api_test.ts` make more sense now. 😄

ryanmurakami:
Thanks for the comments! I'll try out the backprop operations you suggested. I was wondering how that would work as well in this case, which is why I left it out. I'll give it a shot and see what happens.

src/ops_test.ts
ry: Yeah totally understandable line of thinking. Thanks for making the changes!

I'm researching this a bit more and found this inside TF: https://github.com/tensorflow/tensorflow/blob/52dcb2590bb9274262656c958c105cb5e5cc1300/tensorflow/core/framework/op_kernel.h#L714
We may want to attempt to make use of this for efficiency... but I will test your current version once you'd done another pass. Ideally we will see some benchmark improvement with these inplace ops.

src/ops.ts
piscisaureus: This method should documentation.
It's not obvious (other than by reading the source) what the difference with `add()` is.

ryanmurakami: Definitely. I'll add documentation.

ry:
@ryanmurakami any updates on this? From the looks of the current patch it seems you're still poking at it.

ryanmurakami:
@ry  Yep, still messing with it. I made a new debug branch so I won't dirty the pr too much.
At the moment, the `.assign` is giving backprop trouble. It seems to be messing up the `Tape` records and the op goes missing. I'm trying to debug further to figure out exactly what's up with that. 

I'm leaning towards thinking that `.assign` might not be the best way to handle the in-place replacement. Do you have any thoughts on that?

ry:
@ryanmurakami  I believe the problem is that each tensor has an "id" property which is used to track it in backprop. I think assign erroneously is not changing the id currently.

ryanmurakami:
@ry Aha! That was it. Let me clean this up and I'll ping when it's ready to review again. Thanks!

ryanmurakami:
@ry @piscisaureus Alright, ready to review again. Thanks again for all the feedback so far.

One question, should the setting of `id` just be moved to `.assign()`? Is it behavior expected anytime `.assign()` is called?

ryanmurakami:
Wait... used the wrong op in the test. Fixing.

ryanmurakami:
Alright, good to go.

src/tensor.ts
```diff
@@ -181,10 +185,33 @@ export class Tensor implements types.BasicTensor {
     return ops.cast(this, dtype);
   }
 
+  /** Adds two tensors and returns the resulting tensor.
+   *
+   *    import { tensor } from "propel";
+   *    const x = tensor(1);
+   *    const y = tensor(2);
+   *    const result = x.add(y);
```
ry: Thanks for the extra docs : )  It might be more lively to show broadcasting

```
import * as pr from "propel"
tensor([1, 2, 3]).add(100)
```

src/tensor.ts
```diff
@@ -181,10 +185,33 @@ export class Tensor implements types.BasicTensor {
...
+   *
+   *    import { tensor } from "propel";
+   *    const x = tensor(1);
+   *    x.add_(1);
+   */
+  add_(x: types.TensorLike): Tensor {
+    const result = ops.add(this, this.colocate(x));
+
+    this.assign(result);
+    // Should the id setting be moved into .assign() ?
```
ry: Yes, please. That should be a separate commit before this one.

ryanmurakami: PR for moving `id` setting into assign here: #304

ry:
@ryanmurakami I've been thinking there's not much benefit to these in-place ops unless we can reuse memory. To that end I've been toying around with how this might work in the DL backend: https://gist.github.com/ry/6d09c979448ba2ce49530da87f10c30d
It's working for CPU (reusing TypedArrays) but not for WebGL. I haven't looked into the TF backend yet.. (this https://github.com/tensorflow/tensorflow/blob/52dcb2590bb9274262656c958c105cb5e5cc1300/tensorflow/core/framework/op_kernel.h#L714 looked promising.) But WebGL is where we're having the most memory problems and why I was thinking about these in-place ops in the first place - so I think that's the one to get working first.
The other concern is about the backprop: it's safe to delete the forward pass values for addition, because the gradient doesn't depend on the values. However in the case of, say, multiplication, the forward pass values are needed for the backwards pass. I think it could still be a useful optimization, if it errorred out correctly in the case of backproping thru mul_()... I think the thing to do is get the webgl inplace add_ working first, and evaluate from there.

ryanmurakami:
@ry I think I understand your concerns. Let me take a crack at it as well over the weekend and see what I can come up with. The cpu implementation you did looks good, so let me see what I can do with WebGL. Thanks for the good explanation of the problem!

piscisaureus:
I think in-place ops potentially have some benefit for the webgl backend, even if the actual implementation doesn't mutate the input tensor
If an op is performed in-place "conceptually", we can immediately dispose the input tensor and reuse its texture. 
If the garbage collector cleans up the tensor it'll destroy its backing texture, and propel has to create a new one later, which are expensive operations.

ry:
@piscisaureus Good point. So we should be able to measure some performance benefit from with @ryanmurakami's current patch. Maybe something like this:
```javascript
let t = randn([50, 50])
for (let i = 0; i < 1000; i++) {
  let s = randn([50, 50])
  if (inplace) {
    t.add_(s);
  } else { 
    t = t.add(s);
  }
}

**Comment by ryanmurakami** [comment]
I took a stab at trying to do some memory benchmarking with @ry's code above, but I haven't tried to measure memory before and I don't have much faith in the results: https://gist.github.com/ryanmurakami/d9d574692aae8a0bd56b3ef53127a79d

I would love to be able to see a noticeable improvement with this approach before adopting it. Any ideas?

**Comment by qti3e** [comment]
In the case just you want to compare functionalities:
```js
const { randn } = require('../src/api')
const loops = 50000
const shape = [50, 50]

// test inplace op

let t = randn(shape)
let used = process.memoryUsage().heapUsed
let start = Date.now()
for (let i = 0; i < loops; i++) {
  let s = randn(shape)
  t.add_(s)
}
const inplaceHeapSize = process.memoryUsage().heapUsed - used
const inplaceExeTime = Date.now() - start

// test assignment op

t = randn(shape)
used = process.memoryUsage().heapUsed
start = Date.now()
for (let i = 0; i < loops; i++) {
  let s = randn(shape)
  t = t.add(s)
}
const assignmentHeapSize = process.memoryUsage().heapUsed - used
const assignmentExeTime = Date.now() - start

// print results

console.log("inplace op:");
console.log(`   exe time: ${inplaceExeTime / 1000} s`)
console.log(`   heap size:${inplaceHeapSize / (1 << 20)} MB `)

console.log("assignment op:");
console.log(`   exe time: ${assignmentExeTime / 1000} s`)
console.log(`   heap size:${assignmentHeapSize / (1 << 20)} MB `)

```

ryanmurakami:
@qti3e Thanks for putting that together and sharing! I'm curious what you saw when running it? 

After running it multiple times, the results for heap are wildly inconsistent. Most of the heap size outputs are negative, since the after op heap used is less than when it started. My assumption is this is caused by a gc run at some point during the ops. I also tried adding in a manual gc run before each op and it's still pretty inconsistent. 

The exec times are more consistent, and it's leaning towards `add_` taking longer. This makes sense, since the `add_` implementation is basically calling `add` and then `assign`, so there are extra cycles taken for the in-place.

ry:
I haven’t run this yet - but its worth noting that the different backends will probably give quite different result:
The default on the command line will use the TF backend. Setting the env var PROPEL=dl will use the dl cpu backend. Using the webgl backend requires running the bench in a browser.

qti3e:
Om...
@ry Does `process.memoryUsage()` decrease previously returned value from its result?
if so we should just remove `used` vars and everything would be fine.
but otherwise, I think it's due to garbage collection so that it frees up memory and then the second call on `memoryUsage()` would return a smaller value than the first time

ryanmurakami:
So, with the above benchmarks, the `add` result references at `t = t.add(s)` are going to be thrown out and eligible to be garbage collected. With that happening, I don't think we'd see or be able to measure a consistent difference in heap because we're constantly trying to measure before a gc run. If I store that reference in an array that's still in scope when the heap is measured, there's a significant difference in heap used between ops (~30mb). Example:
```js
const results = []
t = randn(shape)
for (let i = 0; i < loops; i++) {
  let s = randn(shape)
  results.push(t.add(s))
}
```

Is this the right approach?

ry:
@ryanmurakami  I'm going to postpone looking into this until I have a few more things implemented. I'm working on a ResNet implementation, and plan spend time optimizing it - I will come back around to this after that bit is complete. Many apologizes for leaving this hanging - I just don't want to introduce potentially destabilizing optimizations before a bit more of the core framework is fleshed out.

ryanmurakami:
@ry Absolutely! Do you just want to leave the PR open, or should we go ahead and close it for now?

ry:
@ryanmurakami let's just leave it open - it's a good start.

ry:
@ryanmurakami I will indeed close this for now, as it's been a few weeks and I'm still unsure about how to move forward here. I will @ mention you when I deal with in-place ops again. Thank you.

Issue #238, opened 2018-02-08
Allowing to write custom descriptions for each paper


qti3e:
Hi
How about adding the ability to write a custom description for each published paper just like GitHub's gists.
It would be great to have keywords and search for keywords too

ry:
Yes, a title would be great

alaq:
Hello, I am working on this issue.

ry:
Thanks @alaq - looking forward to it

piscisaureus:
fixed: #299

PR #299, opened 2018-02-28
notebook title

alaq:
This pull request lets users add titles to notebooks.

- A "title" h1 is added after the notebookBlurb and before the clone button.
- If there is no title, "Untitled Notebook" is displayed in a light grey.
- If the user is logged in, an edit button is displayed.
- Upon clicking on edit, a form is displayed and the user can edit the title. The changes are sent to Firestore.

Addresses #238

### DIFF
--- a/website/db.ts
+++ b/website/db.ts
@@ -42,6 +42,7 @@ export interface UserInfo {
 export interface NotebookDoc {
   cells: string[];
   owner: UserInfo;
+  title: string;
   updated: Date;
   created: Date;
 }
@@ -91,6 +92,7 @@ class DatabaseFB implements Database {
     const docRef = nbCollection.doc(nbId);
     await docRef.update({
       cells: doc.cells,
+      title: doc.title,
       updated: firebase.firestore.FieldValue.serverTimestamp(),
     });
   }
@@ -115,6 +117,7 @@ class DatabaseFB implements Database {
         photoURL: u.photoURL,
         uid: u.uid,
       },
+      title: "",
       updated: firebase.firestore.FieldValue.serverTimestamp(),
     };
     console.log({newDoc});
@@ -248,6 +251,10 @@ export function enableMock(): DatabaseMock {
   return d;
 }
 
+export function ownsDoc(userInfo: UserInfo, doc: NotebookDoc): boolean {
+  return userInfo && userInfo.uid === doc.owner.uid;
+}
+
 function lazyInit() {
   if (db == null) {
     firebase.initializeApp(firebaseConfig);
@@ -300,6 +307,7 @@ grad(f)([-3, -0.5, 0.5, 3])
 const defaultDoc: NotebookDoc = {
   cells: defaultDocCells,
   owner: defaultOwner,
+  title: "Sample Notebook",
   updated: new Date(),
   created: new Date(),
 };
--- a/website/main.scss
+++ b/website/main.scss
@@ -113,12 +113,47 @@ header {
   color: $textColor;
   padding: $size0 0;
 
+  h2 {
+    flex: 1 0 auto;
+    margin-top: 0;
+    margin-bottom: 0;
+    align-self: center;
+  }
+
   p {
     font-size: 18px;
     line-height: 28px;
     font-weight: 100;
     margin-top: -2px;
     margin-bottom: 0;
+  }
+
+  .title-input {
+    flex: 1 0 auto;
+  }
+
+  .untitled {
+    color: #aaa;
+  }
+
+  .title-form {
+    flex: 1 0 auto;
+  }
+
+  button {
+    margin-left: $size1;
+    align-self: center;
+  }
+
+  .title {
+    display: flex;
+    margin-bottom: 1em;
+  }
+
+  input {
+    flex: 1 0 auto;
+    align-self: center;
+    height: $size1;
   }
 }
 
@@ -481,7 +516,7 @@ header {
   margin-top: 50px;
 }
 
-.clone {
+.green-button {
   background: $githubGreen;
   color: #fff;
   border: none;
--- a/website/notebook.ts
+++ b/website/notebook.ts
@@ -506,6 +506,8 @@ export interface NotebookState {
   doc?: db.NotebookDoc;
   errorMsg?: string;
   isCloningInProgress: boolean;
+  editingTitle: boolean;
+  typedTitle: string;
 }
 
 // This defines the Notebook cells component.
@@ -516,6 +518,8 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
       doc: null,
       errorMsg: null,
       isCloningInProgress: false,
+      editingTitle: false,
+      typedTitle: ""
     };
   }
 
@@ -546,6 +550,16 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
       this.update(doc);
     }
   }
+
+  async onSaveTitle(doc) {
+    doc.title = this.state.typedTitle;
+    this.setState({ ...doc, editingTitle: false });
+    this.update(doc);
+   }
+
+   async onTypedTitle(event) {
+     this.setState({ typedTitle: event.target.value });
+   }
 
   async onDelete(i) {
     const doc = this.state.doc;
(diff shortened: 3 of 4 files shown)


website/notebook.ts
```diff
@@ -601,14 +616,71 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
 
     } else {
       const doc = this.state.doc;
+
+      const titleEdit = h(
+        "form",
+        { class: "title", onSubmit: event => this.onSaveTitle(event, doc) },
+        null,
```
ry: What's the point of this null here? I think you can remove the array on the line below, and this null, and just have the h() elements be arguments to the `h("form", ...`

website/notebook.ts
```diff
@@ -601,14 +616,71 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
...
+            "Save"
+          ),
+          h(
+            "button",
+            {
+              class: "edit-title",
+              onClick: () => this.setState({ editingTitle: false })
+            },
+            "Cancel"
+          )
```
ry: nitpicking here: but this html code is taking up too much space, would prefer it condensed like this:
```
          h("button", {
              class: "edit-title",
              onClick: () => this.setState({ editingTitle: false })
          }, "Cancel")
```

website/notebook.ts
```diff
@@ -601,14 +616,71 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
...
+
       const cloneButton = this.props.userInfo == null ? ""
         : h("button", {
             "class": "clone",
             "onClick": () => this.onClone(),
           }, "Clone");
 
       body = [
-        h("header", null, notebookBlurb(doc), cloneButton),
+        h("header", null, notebookBlurb(doc), title, cloneButton),
```
ry: Nice.

website/db.ts
```diff
@@ -300,6 +303,7 @@ grad(f)([-3, -0.5, 0.5, 3])
 const defaultDoc: NotebookDoc = {
   cells: defaultDocCells,
   owner: defaultOwner,
+  title: "",
```
ry: "Sample Notebook"

website/notebook.ts
```diff
@@ -601,14 +616,71 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
...
+      );
+
+      const titleDisplay = h("div", { class: "title" }, [
+        h(
+          "h2",
+          {
+            class: doc.title != null ? "title-input" : "title-input untitled",
+            value: doc.title
+          },
+          `${doc.title != null ? doc.title : "Untitled Notebook"}`
```
ry: Show "untitled notebook" in the case that the title is empty too:
```
doc.title != null && doc.title.length > 0 ? doc.title : "Untitled Notebook"
```

website/notebook_test.ts
```diff
@@ -38,6 +38,14 @@ testBrowser(async function notebook_Notebook() {
   const blurbs = document.getElementsByClassName("blurb");
   assert(1 === blurbs.length);
 
+  // Check that we rendered the title.
+  const title = document.getElementsByClassName("title-input");
+  assert(1 === title.length);
```
ry: Could you also test that `title[0].innerText` is set properly?

website/notebook.ts
```diff
@@ -601,14 +616,71 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
...
+              class: "title-input",
+              onChange: event => this.onTypedTitle(event),
+              value: doc.title,
+            },
+            `${doc.title ? doc.title : ""}`
+          ),
+          h(
+            "button",
+            {
+              class: "edit-title clone",
```
ry: Don't use the clone class here.

website/notebook_test.ts
```diff
@@ -38,6 +38,14 @@ testBrowser(async function notebook_Notebook() {
...
+
+  // Because we aren't logged in, we shouldn't see an edit button for the title.
+  const editButtons = document.getElementsByClassName("button.edit-title");
```
ry: document.getElementsByClassName("edit-title");

website/notebook.ts
```diff
@@ -601,14 +616,71 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
...
+            "Cancel"
+          )
+        ]
+      );
+
+      const titleDisplay = h("div", { class: "title" }, [
+        h(
+          "h2",
+          {
+            class: doc.title != null ? "title-input" : "title-input untitled",
```
ry: Don't give the displayed title the same class as the editable title. This h2 doesn't need a class at all, in fact.

website/notebook_test.ts
```diff
@@ -38,6 +38,14 @@ testBrowser(async function notebook_Notebook() {
   const blurbs = document.getElementsByClassName("blurb");
   assert(1 === blurbs.length);
 
+  // Check that we rendered the title.
```
ry: Very nice. Thanks for adding tests! 

Can you also add a test for when a user is logged in, that when the edit button is clicked, it displays the title entry, and save button, and that the mock database sees an update when save is clicked?

website/notebook.ts
alaq: You are correct, I removed it.

website/notebook.ts
alaq: Thank you, fixed it.

website/main.scss
```diff
@@ -120,6 +120,42 @@ header {
...
+  .title {
+    display: flex;
+    margin-bottom: 1em;
+    height: 40px;
+  }
+
+  input {
+    flex: 1 0 auto;
+    align-self: center;
+    height: 26px;
```
ry: Use $size2 here

website/main.scss
```diff
@@ -120,6 +120,42 @@ header {
...
+
+  button {
+    margin-left: 0.7em;
+    align-self: center;
+  }
+
+  .title {
+    display: flex;
+    margin-bottom: 1em;
+    height: 40px;
```
ry: Use $size2 here or make a $size3: 64px if that's too small.
(Does this element really need a height?)

website/main.scss
```diff
@@ -120,6 +120,42 @@ header {
...
+    color: #aaa;
+  }
+
+  .title-form {
+    flex: 1 0 auto;
+    height: 40px;
+  }
+
+  button {
+    margin-left: 0.7em;
```
ry: Would prefer one of the $size variables here.

website/main.scss
```diff
@@ -120,6 +120,42 @@ header {
...
+    flex: 1 0 auto;
+  }
+
+  .title-input.untitled {
+    color: #aaa;
+  }
+
+  .title-form {
+    flex: 1 0 auto;
+    height: 40px;
```
ry: Would prefer one of the $size variables here. (Does this really need a height?)

website/notebook.ts
alaq: I created the save button class. Should I rename "clone" to "green-button" instead?

ry: Sure, that sounds good.

website/notebook.ts
```diff
@@ -601,14 +616,46 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
...
+              onClick: () => this.setState({ editingTitle: false })
+          }, "Cancel")
+        ]);
+
+      const titleDisplay = h("div", { class: "title" }, [
+        h("h2", {
+          class: doc.title != null ? "" : "untitled",
+          value: doc.title
+        }, doc.title || "Untitled Notebook"),
+        this.props.userInfo ? h("button", {
```
ry: The edit button should only display if you are the owner:
```
this.props.userInfo && this.props.userInfo.uid === doc.owner.uid
```

website/notebook.ts
```diff
@@ -601,14 +616,46 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
...
+                class: "edit-title",
+                onClick: () => this.setState({ editingTitle: true })
+              }, "Edit")
+          : h("p", { class: "" }, "")
```
ry: Instead of `{ class: "" }` put `null` here

website/notebook_test.ts
```diff
@@ -80,3 +89,50 @@ testBrowser(async function notebook_focusNextCell() {
   assert(!cellEls[0].classList.contains("notebook-cell-focus"));
   assert(cellEls[1].classList.contains("notebook-cell-focus"));
 });
+
+testBrowser(async function notebook_NotebookLoggedIn() {
```
ry: Nice. Thanks for adding this test.

website/main.scss
alaq: I changed it but then the input doesn't match the height of the buttons like it does on Github.

website/main.scss
alaq: I removed it, thank you.

website/notebook_test.ts
```diff
@@ -82,3 +91,53 @@ testBrowser(async function notebook_focusNextCell() {
...
+  // Check that we rendered the title.
+  const title = document.querySelectorAll("div.title > h2");
+  assert(1 === title.length);
+  assert("Sample Notebook" === title[0].innerHTML);
+
+  /* TODO Test needs Firebase auth
+  // Because we are logged in, we should see an edit button for the title.
+  const editButton = document.getElementsByClassName("edit-title");
+  assert(1 === editButton.length);
+  */
```
alaq: I believe this test cannot work since `db.active.ownsDoc()` is going to get the Firebase user via `auth.currentUser`.

ry: True. Damn. In that case instead of ownsDoc use this:
```
this.props.userInfo && this.props.userInfo.uid === doc.owner.uid
```

website/main.scss
ry: Hm, exact matches don't matter... But if it looks totally wack, then just do whatever. I'm just trying to avoid having many random values in the CSS.

ry:
Beautiful patch Adrien! Thank you!
I'll upload this to the main site momentarily.

PR #300, opened 2018-03-01
Correct bug jasmine_shim.

ry:


tools/jasmine_shim.ts
```diff
@@ -44,8 +44,12 @@ const matchTesters: MatchTesters = {
       if (a === b) {
         return true;
       }
+      if (a && typeof a === "string" && b && typeof b === "string") {
```
piscisaureus: remove this case

tools/jasmine_shim.ts
```diff
@@ -44,8 +44,12 @@ const matchTesters: MatchTesters = {
...
+        return a === b;
+      }
+      // Note that isNaN(string) is false.
```
piscisaureus: s/false/true/

tools/jasmine_shim.ts
```diff
@@ -44,8 +44,12 @@ const matchTesters: MatchTesters = {
...
       if (isNaN(a) && isNaN(b)) {
-        return true;
```
piscisaureus:
```
if (typeof a === "number" && typeof b === "number" && isNaN(a) && isNaN(b)) {
  return true;
}
```

tools/jasmine_shim_testers.ts
```diff
@@ -59,9 +59,13 @@ export const matchTesters: MatchTesters = {
       if (a === b) {
         return true;
       }
-      if (isNaN(a) && isNaN(b)) {
+      if (typeof a === "number" && typeof b === "number" && isNaN(a) && isNaN(b)) {
         return true;
       }
+      // Note that isNaN(string) is true.
+      if (isNaN(a) && isNaN(b)) {
```
piscisaureus: Yeah so since you changed line 62, you don't have to add this any more.

tools/jasmine_shim_testers.ts
```diff
@@ -0,0 +1,118 @@
+import { inspect } from "util";
+
+export function runTests() {
```
piscisaureus: The name runTests is not appropriate. Should be "generateMethods", rather.

ry: Changed.

Issue #301, opened 2018-03-01
Avatar does not resize on website


cemjotform:
It seems Github does not resize the images based on query parameters anymore. A simple height/width attribute to img dom will solve probably.

![screen shot 2018-03-01 at 21 00 52](https://user-images.githubusercontent.com/28526763/36861406-cda4cdd8-1d94-11e8-840d-0f13d69424f5.png)

PR #302, opened 2018-03-01
common :: added height/width to avatar image

koab:
For the problem that I mentioned at https://github.com/propelml/propel/issues/301


ry:
@koab Thanks for fix! I'll just land this in #303 (no need for cla for such small changes)

PR #311, opened 2018-03-03
[WIP] implement imsave

qti3e:


src/im.ts
```diff
@@ -157,3 +223,26 @@ export async function imread(filename: string, mode: Mode = "RGBA")
   }
   return await webImageDecoder(filename, mode);
 }
+
+export async function imsave(tensor: Tensor, filename: string
+    , handler?: "PNG" | "JPEG"): Promise<void> {
+  if (IS_NODE) {
+    const image = toUint8Image(tensor);
+    if (!handler) {
+      const tmp = filename.split(".");
```
ry: path.extname

src/im.ts
```diff
@@ -157,3 +223,26 @@ export async function imread(filename: string, mode: Mode = "RGBA")
   }
   return await webImageDecoder(filename, mode);
 }
+
+export async function imsave(tensor: Tensor, filename: string
+    , handler?: "PNG" | "JPEG"): Promise<void> {
```
ry: no comma first style. 
```
export async function imsave(tensor: Tensor, 
                             filename: string,
                             handler?: "PNG" | "JPEG"): Promise<void> {
```

src/im.ts
```diff
@@ -122,6 +177,17 @@ function jpegReadHandler(filename: string, mode: Mode): Tensor {
   return toTensor(img.data, img.height, img.width, mode);
 }
 
+function jpegSaveHandler(filename: string, image: Uint8Image): void {
+  console.warn("The JPEG encoder is just working and it may damage your data." +
+               "\nUse it at your own risk.");
```
ry: Not  necessary. This is alpha software - users already know its buggy.

src/im.ts
```diff
@@ -38,6 +44,40 @@ function toTensor(data: Uint8Array, height: number, width: number,
...
 
+function toUint8Image(image: Tensor): Uint8Image {
+  const shape = image.shape;
+  let data: TypedArray;
+  let height: number;
+  let width: number;
+  if (shape.length === 3) {
+    height = shape[0];
+    width = shape[1];
+    const cn = shape[0];
```
ry: Use `channels`, not `cn`

src/im.ts
```diff
@@ -15,11 +15,17 @@
...
 // This module allows Propel to read PNG and JPG images.
 
-import { Tensor } from "./api";
+import { fill, Tensor } from "./api";
 import { convert } from "./tensor";
-import { Mode } from "./types";
+import { Mode, TypedArray } from "./types";
 import { IS_NODE, nodeRequire } from "./util";
 
+type Uint8Image = {
```
ry: I think this is unnecessary

qti3e:
@ry PTAL

src/im.ts
```diff
@@ -15,29 +15,61 @@
...
+    if (channels > 4 || channels === 2) {
+      throw new Error(`${shape} does not match any valid image mode.`);
+    }
+    if (channels === 1) {
+      // Grayscale to RGB
+      image = image.concat(2, data, data);
+    }
+    if (channels <= 3) {
+      // RGB to RGBA
+      image = image.concat(2, fill(255, [1, height, width]));
```
qti3e:
```
Error: cannot compute Concat as input #2 was expected to be a uint8 tensor but is a float tensor
    at OpsTF.concat (/home/qti3e/Projects/propel/src/tf.ts:438:23)
    at exports.concat.inputs (/home/qti3e/Projects/propel/src/ops.ts:428:15)
    at Object.opFunc (/home/qti3e/Projects/propel/src/ops.ts:94:44)
    at Tensor.concat (/home/qti3e/Projects/propel/src/tensor.ts:511:16)
    at Object.toUint8Image (/home/qti3e/Projects/propel/src/im.ts:58:21)
    at im_toUint8ImageRGB (/home/qti3e/Projects/propel/src/im_test.ts:140:20)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:160:7)
```

src/im.ts
```diff
@@ -15,29 +15,61 @@
...
     return tensor;
   }
-  tensor = tensor.slice(0, 3);
+  tensor = tensor.slice(0, [height, width, 3]);
   if (mode === "RGB") {
     return tensor;
   }
   if (mode === "L") {
-    return tensor.reduceMean([0], true);
+    return tensor.reduceMean([2], true);
```
qti3e: I think it's source of `im_toGrayscale`  failing.

src/im.ts
ry: I think the API we'd like to have would be: 
```
fill(255, [1, height, width], { dtype: "uint8" })
````
But that doesn't work at the moment. As a hack you can do
```
fill(255, [1, height, width]).cast("uint8")
```

Comment (qti3e) [reply src/im.ts]: why not using a default dtype for all functions to prevent those things in the future? (I'm not against casting and your proposed API)
if the `Concat` (or any other function) requires a `uint8` so fill should return that.
I'm thinking of a global configuration.

Comment (ry) [reply src/im.ts]: Hm- well float32 is the default. But maybe it makes sense to automatically cast arguments to concat to the same dtype as the first...

File: src/im_test.ts [hunk 8]
```diff
@@ -14,48 +14,143 @@
...
   const data = img.dataSync();
   assert(data.length === 64 * 64);
+
+  // fails
+  // Error: Color mismatch. actual: 84 expected 255
+  // pixel(62, 3) = [255, 255, 255, 255] : proven at im_toRGB & im_pngEncoder
+  // so (expected):
+  // grayscale(62, 3) = [255] : (255 + 255 + 255) / 3
+  // but (actual):
+  // grayscale(62, 3) = [84]
```
qti3e: @ry any idea on this bug?

qti3e: oh! I found it.

ry: Yes, it seems to be due to reduceMean() being used on uint8 tensors
https://gist.github.com/ry/c59021c0e9c6d86f7602c7ebb27a0b24
reduceMean on uint8 is a little suspect... but I think this is also not an expected result. 

For now you can maybe `.cast("int32")`

src/im_test.ts
```diff
@@ -14,48 +14,143 @@
...
+// printColors(imagecreatefrompng(__DIR__ . "/testdata/sample.png"));
+// echo "\n\nJPEG:\n";
+// printColors(imagecreatefromjpeg(__DIR__ . "/testdata/sample.jpg"));
+// ?>
+// [x, y, [r, g, b, a]]
+type SampleData = Array<[number, number, number[]]>;
 // The tests use these files:
 // fa57d083e48e999ed3f210aefd92e5f7  testdata/sample.png
 const pngPath = fetch2ArgManipulation("src/testdata/sample.png");
+const pngData: SampleData = [
```
ry: I'd rather you just check manually as you had before... The reason is there ends up being a lot of code to handle all this custom formats - which may have bugs. Explicitly handling the test cases is easier to read too.

src/im_test.ts
```diff
@@ -14,48 +14,143 @@
...
+//     $b = $rgb['blue'];
+//     echo json_encode([X[$i], Y[$i], [$r, $g, $b, 255]]) . ",\n";
+//   }
+// }
+// echo "PNG:\n";
+// printColors(imagecreatefrompng(__DIR__ . "/testdata/sample.png"));
+// echo "\n\nJPEG:\n";
+// printColors(imagecreatefromjpeg(__DIR__ . "/testdata/sample.jpg"));
+// ?>
+// [x, y, [r, g, b, a]]
```
ry: I'd rather not have this comment as the sample is stored in the repo. But good to know in the PR how it was made.

src/im_test.ts
```diff
@@ -14,48 +14,143 @@
...
 // d37ff170a58223de46152ce837b8e0c4  testdata/sample.jpg
 const jpgPath = fetch2ArgManipulation("src/testdata/sample.jpg");
+const jpegData: SampleData = [
+  [2, 3, [48, 50, 49, 255]],
+  [45, 5, [32, 39, 47, 255]],
+  [62, 3, [48, 49, 51, 255]],
+  [34, 5, [255, 228, 255, 255]]
+];
+
+function assertPixelEqual(data: TypedArray, image: Tensor,
```
ry: assertAllEqual should be sufficient

qti3e:
@ry PTAL

src/im.ts
```diff
@@ -152,3 +209,27 @@ export async function imread(filename: string, mode: Mode = "RGBA")
...
+    const image = toUint8Image(tensor);
+    if (!handler) {
+      const path = nodeRequire("path");
+      const ext = path.extname(filename).toLowerCase();
+      if (ext === ".png" || ext === ".jpeg") {
+        handler = ext.substr(1).toUpperCase();
+      }
+    }
+    handler = handler || "PNG";
+    switch (handler){
```
ry: style

src/im.ts
```diff
@@ -152,3 +209,27 @@ export async function imread(filename: string, mode: Mode = "RGBA")
...
+      case "PNG":
+        return await pngSaveHandler(filename, image);
+      case "JPEG":
+        return jpegSaveHandler(filename, image);
+    }
+    throw new Error(`Unsupported image format "${handler}"`);
+  }
+  // todo: show image instead of saving it
```
ry: Throw an error about imshow not being supported in browsers.

src/im.ts
```diff
@@ -15,27 +15,61 @@
...
+      // RGB to RGBA
+      const dtype = image.dtype;
+      image = image.concat(2, fill(tensor(255, {dtype}), [height, width, 1]));
+    }
+    // RGBA to TypedArray
+    data = image
+      .reshape([height * width * 4])
+      .dataSync();
+  }
+  // todo: support monochrome
```
ry: Let's just do this now. I think reduceMean should be working.

src/im.ts
```diff
@@ -15,27 +15,61 @@
...
+  if (data) {
+    data = Uint8Array.from(data);
```
ry: ClampedUint8Array i think will work better

src/im.ts
```diff
@@ -15,27 +15,61 @@
...
+    return image;
   }
   if (mode === "L") {
-    return tensor.reduceMean([2], true);
+    return image.reduceMean([2], true);
   }
   throw new Error("Unsupported convertion mode.");
 }
 
+export function toUint8Image(image: Tensor) {
```
ry: Could you add

```
interface Image {
  width: number;
  height: number;
  channels: number;
  data: ClampedUint8Array;
}
```

and use that as the return type here. (Although you're not using channels, I want to replace ImshowData from matplotlib with this new Image interface.)

qti3e: I think `channels` is not necessary due to the fact that we always return h*w*4 (RGBA) ClampedUint8Array

src/im.ts
```diff
@@ -15,27 +15,77 @@
...
+      // RGB to RGBA
+      image = image.concat(2, fill(tensor(255, {dtype}), [height, width, 1]));
+    }
+    // Convert to 1D array
+    data = image
+      .reshape([height * width * 4])
+      .dataSync();
+  }
+  if (shape.length === 2) {
+    // convert to a 3D tensor
```
ry: Throw "Unsupported image rank. Use [H, W, 1] for monochrome or [H, W, 3] for RGB."  when image.rank == 2

src/im_test.ts
```diff
@@ -24,38 +25,95 @@ const pngPath = fetch2ArgManipulation("src/testdata/sample.png");
...
+  const img = await imread(pngPath, "L");
+  const rawImage = toUint8Image(img);
+  assert(rawImage.width === 64);
+  assert(rawImage.height === 64);
+  assert(rawImage.data.length === 64 * 64 * 4);
+  const data = rawImage.data;
+  assertAllClose(getPixel(1, data, 2, 3, true), [255]);
+  assertAllClose(getPixel(1, data, 20, 20, true), [237]);
+  assertAllClose(getPixel(1, data, 62, 3, true), [255]);
+  assertAllClose(getPixel(1, data, 34, 5, true), [129]);
```
ry: Nice tests!

src/im.ts
qti3e: the current `imshow` function accepts 2Ds and I think you're going to this function to reimplement that, won't not-supporting 2Ds break other previous tests?

ry: Ok, good point. But the code in this branch is repeated above... Can you "DRY" that up?

qti3e: done

PR #312, opened 2018-03-03
Add align to tslint.json and fix errors.

ry:


tools/test_browser.ts
```diff
@@ -72,17 +72,17 @@ if (!filterExpr) {
...
     timeout: 10 * 1000,
   },
-  {
+                        {
     path: "notebook/",
     doneMsg: /Propel onload/,
     timeout: 10 * 1000,
   },
-  {
+                        {
```
piscisaureus: This looks pretty wonky...

ry:
Dang - I totally missed that.

tools/test_browser.ts
ry: Dang - I totally missed that.

ry: Fixed. PTAL.

Issue #315, opened 2018-03-04
Creating new document cause error


qti3e:
```
Error
Function DocumentReference.update() called with invalid data. Unsupported field value: undefined (found in field title)
```

alaq:
#318 will address this.

qti3e:
Closed in favor of #318 :)

PR #318, opened 2018-03-04
fix title error on new notebook

alaq:
Running a new notebook returned an error as there was no title on the new notebooks.

Addresses #315

### DIFF
--- a/website/db.ts
+++ b/website/db.ts
@@ -92,7 +92,7 @@ class DatabaseFB implements Database {
     const docRef = nbCollection.doc(nbId);
     await docRef.update({
       cells: doc.cells,
-      title: doc.title,
+      title: doc.title || "",
       updated: firebase.firestore.FieldValue.serverTimestamp(),
     });
   }
@@ -138,6 +138,7 @@ class DatabaseFB implements Database {
         photoURL: u.photoURL,
         uid: u.uid,
       },
+      title: "",
       updated: firebase.firestore.FieldValue.serverTimestamp(),
     };
     console.log({newDoc});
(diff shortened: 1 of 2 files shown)


ry:
Thanks Adrien! Could we get a test for this to prevent the error in the future?

website/db.ts
```diff
@@ -92,7 +92,7 @@ class DatabaseFB implements Database {
     const docRef = nbCollection.doc(nbId);
     await docRef.update({
       cells: doc.cells,
-      title: doc.title,
+      title: doc.title || "",
```
qti3e: Shouldn't it be `Untitled notebook`?

alaq:
Will add the test now.

website/db.ts
alaq: If the title is "", it gets replaced with a grey "Untitled Notebook". And we want to display something, even when the title is empty.

ry:
@alaq Actually I realize now that we don't have any tests that run against the actual database (only the mock database), which is required to properly test this error. I've added an issue #321 to address this. 

Please squash this commits and I will land them as is.

website/notebook.ts
```diff
@@ -610,7 +610,7 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
 
       const titleDisplay = h("div", { class: "title" }, [
         h("h2", {
-          class: doc.title != null ? "" : "untitled",
+          class: doc.title && doc.title.length ? "" : "untitled",
```
alaq: I added this check here since we also want the title greyed out if `doc.title === ""`.

PR #322, opened 2018-03-05
Parcel upgrade to 1.6.x

### LINKED ISSUE #264: Upgrade parcel to 1.6

ryanmurakami:
Fixes #264
In Parcel 1.6x, there were some changes made where `async` functions are transformed to `regeneratorRuntime`, which is `undefined`. Used this issue to find a resolution: https://github.com/parcel-bundler/parcel/issues/871

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,9 @@
+{
+  "plugins": [
+    ["transform-runtime",
+      {
+        "polyfill": false,
+        "regenerator": true
+      }]
+  ]
+}
--- a/.babelrc
+++ b/.babelrc
@@ -1,9 +1,0 @@
-{
-  "plugins": [
-    ["transform-runtime",
-      {
-        "polyfill": false,
-        "regenerator": true
-      }]
-  ]
-}
--- a/package.json
+++ b/package.json
@@ -18,6 +18,8 @@
     "@types/node": "^8.9.4",
     "@types/seedrandom": "^2.4.27",
     "acorn": "^5.5.0",
+    "babel-plugin-transform-runtime": "^6.23.0",
+    "babel-runtime": "^6.26.0",
     "codemirror": "^5.35.0",
     "d3": "^4.13.0",
     "http-server": "^0.10.0",
@@ -42,5 +44,6 @@
     "tslint": "^5.9.1",
     "typescript": "^2.7.2",
     "yauzl": "^2.9.1"
-  }
+  },
+  "dependencies": {}
 }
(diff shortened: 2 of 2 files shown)


ryanmurakami:
Looks like the review just failed since [#4](https://github.com/propelml/propel_deps/pull/4) hasn't landed yet.

ry:
@ryanmurakami Awesome! Thanks! 

I'm not sure I understood the babel issue in Parcel - this is a bug on their end? Maybe we should change the browserslist instead, as Devon suggested. We aren't trying to support IE, for example.

ryanmurakami:
@ry Great, yeah I was able to use `browserslist` instead. It isn't documented yet in parcel, so I wasn't quite sure how it should've been used. Being set at `>5%` indicates that browser versions with greater than 5% market share will be transformed for. Specifically, it handles the `async` issue I was dealing with. 

Thanks for the suggestion!

piscisaureus:
Starting with parcel v1.6 it'll run all generated code through babel.
I think this is unnecessary to do at all - typescript can output js in flavors we need.
Is it possible to turn off babel entirely?

ryanmurakami:
It looks like there isn't a way to disable babel. The [typescript parser](https://github.com/parcel-bundler/parcel/blob/master/src/assets/TypeScriptAsset.js) extends the js stuff, which uses babel by default. There might a way to short circuit babel by doing something weird to the babel configuration, but I wasn't able to figure out a way to do it.

ry:
@piscisaureus I agree it would be nice parcel didn't depend on babel - but given that we already have it in our deps (due to nb_transpiler.ts) - I'm in favor of landing this.

piscisaureus:
@ry 
1. nb_transpiler doesn't use babel.
2. up to version 1.5 parcel didn't use babel. It may have been listed as a dependency but it wasn't used anywhere in our code or build process.

piscisaureus:
I'd prefer not to upgrade for now, and see if they provide an option to disable babel transforms on typescript code. We can always revisit after two months.

piscisaureus:
Filed an issue: https://github.com/parcel-bundler/parcel/issues/954

piscisaureus:
Ok, let's land this as a workaround. Thanks @ryanmurakami!
now if only i knew how to make appveyor test this PR...

piscisaureus:
This whole deps situation must have been pretty painful to deal with.

ryanmurakami:
@piscisaureus deps wasn't too bad. Git submodules are always confusing in the beginning. Just for reference, here's the deps PR: https://github.com/propelml/propel_deps/pull/4

ryanmurakami:
Trying to trigger the builds again.

piscisaureus:
Landed in #342

PR #327, opened 2018-03-06
Support @include and @mixin in stylelint.

ry:


piscisaureus:
What does this do? There are no code changes.

ry:
This allows the use of @mixin and @include in scss files. Changes are coming that will use it (just want to land it now)

PR #331, opened 2018-03-06
Add conv2d

ry:


src/conv_test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+// Test cases ported from tensorflow/python/kernel_tests/conv_ops_test.py
+// Dilation tests skipped entirely for now.
+// The naming conventions from the TF test case are followed. However
+// tensorInSizes would be better called inputShape and filterInSizes would be
+// better called filterShape.
+
+interface ConvExample {
+  tensorInSizes: [number, number, number, number]; // NHWC
+  filterInSizes: [number, number, number, number]; // H W InChans OutChans
```
piscisaureus: number[4]

src/conv_test.ts
```diff
@@ -0,0 +1,137 @@
...
+      assertAllClose(actual.dataSync(), ex.expected);
+    }
+  }
+}
+
+function setupValues(format: ImageFormat, dtype: DType,
+                     ex: ConvExample): Tensor {
+  const x1 = rangeJS(1, prod(ex.tensorInSizes) + 1);
+  const x2 = rangeJS(1, prod(ex.filterInSizes) + 1);
+  let input = tensor(x1, { dtype }).reshape(ex.tensorInSizes);
```
piscisaureus: Why isn't this `range(1, prod(ex.tensorInSizes)).cast(dtype)` ?
Now you have to add this rangeJS function.

src/conv_test.ts
ry: That would be nice.
```
src/conv_test.ts (28,25): Property '4' does not exist on type 'Number'. (2339)
```

src/conv_test.ts
ry: Fixed.

src/tensor.ts
```diff
@@ -707,6 +707,33 @@ export class Tensor implements types.BasicTensor {
     }
     return t;
   }
+
+  /** Preforms 2d convolution.
```
piscisaureus: Performs

src/tf.ts
```diff
@@ -501,4 +501,45 @@ export class OpsTF implements types.BackendOps {
...
+                    [inputShapeT, filter, gradient],
+                    convAttrs(opts));
+  }
+}
+
+function convAttrs(opts: types.ConvOpts): AttrDef[] {
+  let strides = [1, 1, 1, 1];
+  if (typeof opts.strides  === "number") {
+    strides = [1, opts.strides, opts.strides, 1];
+  } else if (strides instanceof Array) {
```
piscisaureus: This is always true. You probably mean `opts.strides instanceof`

ry: good catch. fixed.

src/tensor.ts
ry: fixed.

src/tf.ts
```diff
@@ -501,4 +501,45 @@ export class OpsTF implements types.BackendOps {
...
+                      filter: TensorTF, opts: types.ConvOpts): TensorTF {
+    const inputShapeT = int32Small(inputShape);
+    return execute0("Conv2DBackpropInput",
+                    [inputShapeT, filter, gradient],
+                    convAttrs(opts));
+  }
+}
+
+function convAttrs(opts: types.ConvOpts): AttrDef[] {
+  let strides = [1, 1, 1, 1];
```
piscisaureus: what's this initializer for?

ry: removed.

PR #333, opened 2018-03-07
Allow for anonymous notebooks

### LINKED ISSUE #265: "Create Empty Notebook" button doesn't work when not signed in.

ryanmurakami:
Fixes #265 

Creates anonymous notebooks when a user isn't signed in. Any changes are not saved, but the notebook is fully functional in the browser.

Here's what it looks like:
screen shot 2018-03-06 at 9 47 16 pm

The messaging could be better, I'm just overloading the title. I also have not been able to test the experience if the user signs in while the anonymous notebooks is open. It would be nice to save it at that time.

The profile pic can be changed, but it's one of my dog so there won't be any copyright issues. 🐶


ryanmurakami:
Ach, I'm a rebase noob. Opening a different pr without the commit mistakes. Reopened as #335

PR #341, opened 2018-03-07
 [Fix #339] post uint8array instead of tensor to imshow

### LINKED ISSUE #339: outputHandler can not post canvas to main window
```
Uncaught (in promise) DOMException: Failed to execute 'postMessage' on 'Window': #<WeakMap> could not be cloned.
```

qti3e:
image

### DIFF
--- a/src/im.ts
+++ b/src/im.ts
@@ -103,9 +103,8 @@ function webImageDecoder(filename: string, mode: Mode)
   });
 }
 
-export function createCanvas(image: Tensor) {
-  const uint8Image = toUint8Image(image);
-  const { height, width, data } = uint8Image;
+export function createCanvas(image: Image) {
+  const { height, width, data } = image;
   const canvas = document.createElement("canvas");
   canvas.height = height;
   canvas.width = width;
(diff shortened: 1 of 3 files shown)


qti3e:
I guess we can write a test in nb_test.ts to determine if is there any canvas in the document or not (after calling `imshow`)

piscisaureus:
In nb_test, add along these lines:

```ts
testBrowser(async function notebook_fub4rba5() {
  resetPage();
  await new Promise((resolve) => {
    const el = h(nb.Notebook, {
      onReady: resolve,
      code: "imshow('foo/bar');"
    });
    render(el, document.body);
  });
  // Check that a canvas was created in the output div.
  ...
});
```

qti3e:
@piscisaureus I tried this and it failed:
```js
testBrowser(async function notebook_imshow() {
  resetPage();
  await new Promise((resolve) => {
    const el = h(nb.Notebook, {
      nbId: "default",
      onReady: resolve,
      code: `import * as pr from "propel"
      const img = pr.zeros([10, 10, 3])
      pr.imshow(img)`
    });
    render(el, document.body);
  });
  const canvas = document.querySelector(".output-container > .output > canvas");
  assert(canvas !== null);
});
```
I guess it's because nb's ready event does not mean code has been executed and hence there is no canvas in the document.

PR #344, opened 2018-03-08
Add maxPool, moments, stopGradient, sin, cos, tan, sqrt, pow, and new layers api with batchNorm, conv2d, linear

ry:


src/ops.ts
```diff
@@ -276,6 +276,21 @@ defBW("square", (g, x) => {
   return g.mul(x.mul(two));
 });
 
+export const pow = defFW("pow", (x: types.Storage, e: types.Storage) => {
+  return bo.pow(x, e);
+});
+defBW("pow", (g: Tensor) => {
+  throw Error("Not Implemented.");
```
piscisaureus: g.mul(e).mul(x.pow(e.sub(1))) I think

src/dl/math/backends/webgl/binaryop_gpu.ts
```diff
@@ -28,8 +28,9 @@ export const SUB = "return a - b;";
 export const MUL = "return a * b;";
 export const DIV = "return a / b;";
 export const POW = `
-  return (round(mod(b, 2.0)) == 0 || round(mod(b, 2.0)) == 2) ?
-      pow(abs(a), b) : sign(a) * pow(abs(a), b);
+  return (a >= 0.) ? pow(a, b) :
+    ((round(mod(b, 2.0)) == 0 || round(mod(b, 2.0)) == 2) ?
```
piscisaureus: I don't think this works properly when `a < 0` and b is not a round number.

src/experiment.ts
```diff
@@ -169,7 +169,9 @@ export abstract class Experiment {
     gc((keep) => {
       const gradFn = gradParams(lossFn);
       // Forward/Backward pass
+      (this.currentParams as any).isTraining = true;
```
piscisaureus: This cast shouldn't be necessary.

src/experiment.ts
```diff
@@ -169,7 +169,9 @@ export abstract class Experiment {
...
       const gradsAndLoss = gradFn(this.currentParams);
+      (this.currentParams as any).isTraining = false;
```
piscisaureus: ditto

src/ops.ts
ry: thanks. fixed.

tools/gendoc.ts
```diff
@@ -188,6 +188,10 @@ export function genJSON(): DocEntry[] {
       // TODO Ignoring for now.
       log("- ObjectLiteralExpression");
 
+    } else if (ts.isTypeLiteralNode(node)) {
+      // TODO Ignoring for now.
+      log("- TypeLiteral");
```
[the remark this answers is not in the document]
ry: Not sure exactly.. : )

src/tensor.ts
```diff
@@ -388,6 +388,27 @@ export class Tensor implements types.Storage {
     return ops.reduceLogSumExp(this, axes, keepDims);
   }
 
+  /** Prevents from backproping thru the tensor. Returned tensor uses the same
+   * storage.
+   */
+  stopGradient(): Tensor {
```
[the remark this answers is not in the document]
ry: So, those other methods are internal ops - where I think Grad is shorter and consistent with previous gradient ops like reluGrad. This is meant to be familiar to TF users https://www.tensorflow.org/api_docs/python/tf/stop_gradient

src/api.ts
```diff
@@ -183,3 +184,29 @@ export function int32(t: types.TensorLike): Tensor {
 export function float32(t: types.TensorLike): Tensor {
   return tensor(t, { dtype: "float32" });
 }
+
+/** Performs 2d convolution.
+ * The input and filter tensors should both be rank 4 and float32, with the
+ * input formatted as [batch, height, width, channels] and the filter
+ * [height, width, in chans, out chans].
+ *
+ * Additional arguments are ConvOpts { stride, padding, bias }.
```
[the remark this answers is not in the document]
ry: Fixed

src/dl/math/backends/webgl/binaryop_gpu.ts
ry: Yeah, I'm not sure on which real numbers this is defined, at least a = -1 and b = 1/2 is imaginary ... I'd leave it as until we encounter counter-examples

src/experiment.ts
ry: Fixed.

src/experiment.ts
ry: Fixed.

PR #346, opened 2018-03-09
Adds ./tools/dev_website for fast incremental builds 

ry:
The deps commit includes some hacks on parcel to fix a bug - I will send upstream soon

https://github.com/propelml/propel_deps/commit/41035369fe03732c3c8eb08cdff617d46dc11caf#diff-5e9e9af54dc82093dc2e5334dfff25c3

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,19 +89,21 @@ Propel has two major components:
 Due to the notebook and documentation, the website is fairly non-trivial. It
 is written in React (actually Preact), but does not use JSX.
 
-When working on the website, it's often nice to only run the website tests,
-and not worry about the framework. To do this run a local website:
+When working on the website, it's often nice to run a dev server
 
-    ./tools/http_server.js
+    ./tools/dev_website.js
 
-Then to automatically build the website tests on changes, run
+This will automatically rebuild the website on any scss or javascript
+changes.
+
+Sometimes you'll want to work on the tests - this can be done like so:
 
     ./tools/parcel.js tools/test_website.ts -d build/website/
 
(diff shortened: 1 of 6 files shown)


piscisaureus:
×  C:\Users\BertBelder\d\propel3\website\main.scss:undefined:undefined: Missing binding C:\Users\BertBelder\d\propel3\deps\node_modules\node-sass\vendor\win32-x64-57\binding.node
Node Sass could not find a binding for your current environment: Windows 64-bit with Node.js 8.x

Found bindings for the following environments:
  - OS X 64-bit with Node.js 8.x

piscisaureus:
Nice. Although I am a bit concerned by the code duplication between build_website and dev_website.
Also note that the next time someone runs npm update your parcel hacks will be overwritten - this is going to be a pain unless you manage to upstream the pacel patch soon.

ry:
Patch sent upstream: https://github.com/parcel-bundler/parcel/pull/974

tools/build.js
```diff
@@ -6,6 +6,7 @@ if (process.argv.indexOf("clean") >= 0) {
 }
 
 (async() => {
+  run.sh("npm rebuild");
```
piscisaureus: This fails on windows. Cannot spawn .cmd files. Use execSync instead.

piscisaureus: Actually can we not have this here at all? npm rebuild is way too slow for me to want to do that all the time.

ry:
@piscisaureus Removed node-sass and used a symlink to sass. Removed the npm rebuild too. PTAL.

PR #347, opened 2018-03-09
CSS improvements

qti3e:
CSS:
- set a gray-white background for output canvas.
- add styles to edit title input element


qti3e:
I do the same thing as you're saying, tale attention to `cors=false ` section in arguments list : )

src/im.ts
```diff
@@ -96,9 +97,19 @@ function webImageDecoder(filename: string, mode: Mode)
...
+        console.info(`Trying to load image from CORS proxy...`);
+        img.src = "https://cors-anywhere.herokuapp.com/" + filename;
+      } else {
+        img.src = filename;
+      }
     } catch (e) {
-      resolve();
+      if (!cors) {
+        webImageDecoder(filename, mode, true).then(resolve);
+      }else {
```
ry: style

website/main.scss
```diff
@@ -144,6 +144,13 @@ header {
 
   .title-input {
     flex: 1 0 auto;
+    outline: none;
+    border: none;
+    height: 32px;
+    line-height: 32px;
+    border-bottom: solid 1px rgba(0, 0, 0, 0.5);
+
+    @include shadowHover;
```
ry: Looks nice. Could you use the $size variables instead of pixels?

website/main.scss
```diff
@@ -700,6 +707,30 @@ img.avatar {
...
+    linear-gradient(
+      45deg,
+      #efefef 25%,
+      transparent 25%,
+      transparent 75%,
+      #efefef 75%,
+      #efefef
+    );
+  background-position: 0 0, 10px 10px;
+  background-size: 21px 21px;
```
ry: I wasn't able to notice these changes (generally I'd like to stay away from complex gradients like this until we're further along and can think about the design more cohesively)

src/im.ts
```diff
@@ -75,7 +76,7 @@ export function toUint8Image(image: Tensor): Image {
   throw new Error(`Unsupported image rank.`);
 }
 
-function webImageDecoder(filename: string, mode: Mode)
+function webImageDecoder(filename: string, mode: Mode, cors = false)
```
ry: It doesn't seem like there's a way to actually use the cors flag tho, since it's not surfaced at `imread()`.

qti3e: it's a argument to block a potentially infinite loop

website/main.scss
qti3e: I used it instead of loading an image

website/main.scss
qti3e: ok

website/main.scss
```diff
@@ -799,6 +806,30 @@ img.avatar {
   }
 }
 
+.output canvas {
```
ry: Can you add a comment about what this does?

qti3e: done

website/main.scss
yungbeto: styling lgtm 
can you share a quick screen grab of how it looks behind an image? @qti3e

qti3e: image

yungbeto: nice, so an image with some transparency would look something like this? -
image
@qti3e

qti3e: @yungbeto yeah exactly

ry:
LGTM included in #370

PR #359, opened 2018-03-12
Prettier

ry:


tools/package.js
```diff
@@ -33,9 +33,11 @@ function createPackageJson(src, dst, packageJson = {}) {
 
 function checkNoTars(dir) {
   for (const f of fs.readdirSync(dir)) {
-    if (f.match(/\.tgz$/) != null ||
-        f.match(/\.tar.gz$/) != null ||
-        f.match(/\.zip$/) != null) {
+    if (
+      f.match(/\.tgz$/) != null ||
```
ryanmurakami: These can be `!==`

tools/test_browser.ts
```diff
@@ -49,7 +49,7 @@ const debug = !!process.env.PP_TEST_DEBUG;
 const testdl = !!process.env.PP_TEST_DL;
 
 // Run headless only if CI env var is set.
-const headless = (process.env.CI != null);
+const headless = process.env.CI != null;
```
ryanmurakami: I'm assuming this `!=` is to also check for `undefined`, but just checking.

ry:
@piscisaureus has some concerns about prettier still (regarding the formatting of matricies) - will close this for now and do a new patch at some point.

(Thanks for the review @ryanmurakami - I will check this when I do another prettier patch and make sure to address your comments)

PR #360, opened 2018-03-14
[WIP] Improve notebook tests.

ry:

PR #363, opened 2018-03-14
notebook: show accurate stack traces in the notebook

piscisaureus:
There is one TODO item left: when acorn encounters a parse error it
throws a SyntaxError, which code position isn't mapped to the
untranspiled source code.

### DIFF
--- a/website/nb_transpiler.ts
+++ b/website/nb_transpiler.ts
@@ -12,13 +12,10 @@
    See the License for the specific language governing permissions and
    limitations under the License.
  */
+import { AbstractToken } from "acorn";
 import * as acorn from "acorn/dist/acorn";
 import * as walk from "acorn/dist/walk";
 import { assert } from "../src/util";
-
-const importFn = "__import";
-const globalVar = "__global";
-const parseOptions = { ecmaVersion: 8, allowImportExportEverywhere: true };
 
 function noop() {}
 
@@ -39,48 +36,6 @@ function walkRecursiveWithAncestors(node, state, visitors) {
   return walk.recursive(node, state, wrappedVisitors);
 }
 
-class SourceEditor {
-  // TODO: track transformations to enable source maps.
-  parts: string[];
-
-  constructor(text) {
-    this.parts = text.split("");
-  }
-
-  text() {
-    return this.parts.join("");
-  }
-
-  stratify() {
-    const text = this.text();
-    this.parts = text.split("");
-    return text;
-  }
-
-  replace(start, end, str) {
-    let before = "";
-    for (let i = start; i < end; i++) {
-      before += this.parts[i];
-      this.parts[i] = "";
-    }
-
-    if (start === end) {
-      str += this.parts[start];
-    }
-
-    this.parts[start] = str;
-    return before;
-  }
-
-  prepend(node, str) {
-    this.parts[node.start] = str + this.parts[node.start];
-  }
-
-  append(node, str) {
-    this.parts[node.end - 1] += str;
-  }
-}
-
 /* tslint:disable:object-literal-sort-keys*/
 
 const importVisitors = {
@@ -95,10 +50,10 @@ const importVisitors = {
         state.edit.replace(cur.end, spec[i].start, ",");
         cur = spec[i];
       }
-      state.edit.replace(cur.end, src.start, `} = {_:await ${importFn}(`);
+      state.edit.replace(cur.end, src.start, `} = {_:await __import(`);
       state.edit.replace(src.end, node.end, ")};");
     } else {
-      state.edit.replace(node.start, src.start, `await ${importFn}(`);
+      state.edit.replace(node.start, src.start, `await __import(`);
       state.edit.replace(src.end, node.end, ");");
     }
 
@@ -106,17 +61,17 @@ const importVisitors = {
   },
 
   ImportSpecifier(node, state, c) {
-    state.edit.prepend(node, "_:{");
+    state.edit.insertBefore(node, "_:{");
     if (node.local.start > node.imported.end) {
       state.edit.replace(node.imported.end, node.local.start, ":");
     }
-    state.edit.append(node, "}");
+    state.edit.insertAfter(node, "}");
     walk.base.ImportSpecifier(node, state, c);
   },
 
   ImportDefaultSpecifier(node, state, c) {
-    state.edit.prepend(node.local, "_:{default:");
-    state.edit.append(node.local, "}");
+    state.edit.insertBefore(node.local, "_:{default:");
+    state.edit.insertAfter(node.local, "}");
     walk.base.ImportDefaultSpecifier(node, state, c);
   },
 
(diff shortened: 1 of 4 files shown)


website/nb_transpiler.ts
```diff
@@ -256,57 +210,350 @@ const evalScopeVisitors = {
...
+}
 
-  src = edit.text();
+interface Position {
+  line?: number;
+  column?: number;
+}
 
-  return src;
+class MappedChar implements Position {
```
ry: interface?

website/sandbox.ts
```diff
@@ -110,3 +110,10 @@ matplotlib.setOutputHandler({
     rpc.call("imshow", guessCellId(), data);
   }
 });
+
+window.addEventListener("error", (ev: ErrorEvent) => {
+  const error = ev.error;
+  const cellId = guessCellId(error);
+  const stack = transpiler.formatError(error);
+  rpc.call("console", cellId, stack);
+});
```
ry: What causes this error event? Is this part of the iframe API?

website/nb_transpiler.ts
```diff
@@ -256,57 +210,350 @@ const evalScopeVisitors = {
...
+
+  insertBefore({ start }, str: MappedStringLike): void {
+    const mstr = MappedString.convert(str, this.index[start][0]);
+    this.index[start] = mstr.concat(this.index[start]);
+  }
+
+  insertAfter({ end }, str: MappedStringLike): void {
+    const mstr = MappedString.convert(str, this.index[end][0]);
+    this.index[end - 1] = this.index[end - 1].concat(mstr);
+  }
```
ry: I think completed types would be helpful here: insertAfter, insertBefore, replace

website/nb_transpiler.ts
```diff
@@ -256,57 +210,350 @@ const evalScopeVisitors = {
...
+    } else if (str instanceof Array) {
+      return new MappedString(str);
+    } else {
+      return new MappedString(
+        Array.from("" + str).map(char => new MappedChar(char, pos))
+      );
+    }
+  }
+
+  constructor(chars: number | string | MappedChar[] = [], pos?: Position) {
```
ry: why take a number here?

website/nb_transpiler.ts
piscisaureus: Can an interface have a constructor?

website/nb_transpiler.ts
piscisaureus: Because methods inherited from Array will call the constructor with a numerical argument (which is the length it thinks the new Array should have). 
I'm not using the "new array length" here because I don't want the MappedString to have a bunch of undefined elements.

piscisaureus: Because methods inherited from Array will call the constructor with a numerical argument (which is the length it thinks the new Array should have).
I'm not using the "new array length" here because I don't want the MappedString to have a bunch of undefined elements.

website/nb_transpiler.ts
piscisaureus: The first argument is an acorn AST node. Unfortunately we don't have types for acorn; i'll see if I can find them on npm.

ry: https://www.npmjs.com/package/@types/acorn

website/sandbox.ts
piscisaureus: window.onerror received "uncaught" exceptions. Let's say you did this in a notebook:

```
setTimeout(() => {
  throw new Error('err')
}, 100);
```

Currently the error would be swallowed. This makes sure it gets shown to the user; we're guessing the cell originating as well as we can.

I see there is a bug in it though, will fix soon.

website/nb_transpiler.ts
```diff
@@ -256,57 +210,350 @@ const evalScopeVisitors = {
...
-    body,
-    {
+  // Transpiles a repl cell into an async function expression.
+  // The returning string has the form:
+  //   (async (global, import) => {
+  //     ... cell statements
+  //     return last_expression_result;
+  //   })
+  transpile(code: string, name: string): string {
+    const id = `${++this.counter}`;
```
ry: What is counter counting? Is this "id" supposed to be the same as cell ids? Consider adding a comment here.

website/nb_transpiler.ts
```diff
@@ -256,57 +210,350 @@ const evalScopeVisitors = {
...
+    // Store the transpilation result so we can map stack traces to their
+    // untranspiled counterparts.
+    this.originalSources[id] = original;
+    this.transpiledSources[id] = transpiled;
+
+    return transpiled.toString();
+  }
+
+  private static sourcePositionRe =
+    /(?:__transpiled_source_(\d+)__)(?::(\d+))?(?::(\d+))?/;
```
ry: since you're using \d+ here to match the id, why not explicitly make the id a number instead of a string. I'm talking about this line above
```
const id = `${++this.counter}`;
```
maybe make it 
```
const id: number = ++this.counter;
```

website/sandbox.ts
ry: Maybe add a comment about this here with a TODO.

website/sandbox.ts
```diff
@@ -35,21 +35,22 @@ async function importModule(target) {
 
 let lastExecutedCellId: number = null;
 
+const transpiler = new Transpiler();
+
 const rpc = new SandboxRPC(window.parent, {
   async runCell(source: string, cellId: number): Promise<void> {
```
ry: FYI in #360,  I change `cellId` to be a string instead of a number. It doesn't seem like your code makes any hard assumptions about here tho, so it shouldn't be a problem.

website/nb_transpiler_test.ts
```diff
@@ -13,81 +13,177 @@
...
+  t("badfn()", "test1", "test1:1:1");
+  t("var asdfg=42;badfn()", "test2", "test2:1:14");
+  t("\n" +
+    "import * as foo from 'bar';badfn()",
+    "test3", "test3:2:28");
+  t("function wierd() {\n" +
+    "  return badfn();\n" +
+    "} 1+2+3+wierd()",
+    "test4", "test4:2:3", "test4:3:9");
+  t("wierd()", "test5", "test4:2:3", "test5:1:1");
```
ry: Nice tests.

website/nb_transpiler_test.ts
```diff
@@ -13,81 +13,177 @@
...
+      "//# sourceURL=__transpiled_source_1__"
+    ].join("\n");
+
+    const transpiler = new Transpiler();
+    const transpiled = transpiler.transpile(src, "test");
+
+    assert(transpiled === expected,
+          "unexpected transpiled source code\n" +
+          `actual:   ${JSON.stringify(transpiled)}\n` +
+          `expected: ${JSON.stringify(expected)}`
```
ry: In the future we should improve our util.assertEqual to handle this more generally.

website/sandbox.ts
piscisaureus: I fixed the bug

website/nb_transpiler.ts
piscisaureus: I added a comment.

website/nb_transpiler.ts
piscisaureus: It's not the same as the cell id. It's just a number that is incremented every time the transpiler transpiles new code. This allows us look up "old" transpiled code that appears in a stack trace, e.g. when one cell defines a function and then another cell calls that function.

I changed the type to number. (the reason I had it as a string was actually to keep types consistent -- when the id is found in a stack trace using a regexp, it's type will be a string).
But if I change it to number it just works fine.

website/nb_transpiler.ts
piscisaureus: done

website/sandbox.ts
piscisaureus: ok

website/nb_transpiler_test.ts
piscisaureus: thanks

PR #381, opened 2018-03-17
notebook: refactor sandbox rpc channel

piscisaureus:


piscisaureus:
@ry I have some cursory testing of the websocket channel going on in https://github.com/propelml/propel/pull/381/commits/11f0f928ea457851dba7fa0f2383baa6a0858cdf.
I am struggling a bit with testing the iframe channel a bit. It looks like we'll need to add yet another bundle for this, otherwise I see no way to load the tests into an iframe.

ry:
Hmm I don’t have any immediate suggestions - but I think the node tests are good for now. Looks good once ci passes.

PR #382, opened 2018-03-17
website_render fixes

piscisaureus:


tools/dev_website.js
```diff
@@ -62,8 +62,9 @@ async function bundler(build) {
 }
 
 const port = 8080;
-async function devWebsiteServer(build) {
+async function devWebsiteServer(build = false) {
   const b = await bundler(build);
+  if (build) await b.bundle();
```
ry: `serve` should already build it

ry: discussed offline.

PR #387, opened 2018-03-17
dataset: implement naive shuffling

indutny:
cc @ry

### DIFF
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -158,7 +158,7 @@ class SliceDataset extends Dataset {
 
 // TODO
 function stack(tensors: Tensor[], axis = 0): Tensor {
-  throw Error("not implemented");
+  return tensors[0].concat(axis, tensors[1], ...tensors.slice(2));
 }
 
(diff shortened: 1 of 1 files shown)


indutny:
@ry perhaps a bit later :wink:

src/dataset.ts
```diff
@@ -158,7 +158,7 @@ class SliceDataset extends Dataset {
 
 // TODO
 function stack(tensors: Tensor[], axis = 0): Tensor {
-  throw Error("not implemented");
+  return tensors[0].concat(axis, tensors[1], ...tensors.slice(2));
```
ry: Leave as unimplemented - doesn't seem like you're using it

indutny: I'm not, but `BatchDataset` indirectly does.

ry: also it's incorrect : ) it needs expandDim. and it should be in tensor.ts

ry:
Will land in #398 - Thanks Fedor!!

PR #390, opened 2018-03-17
Add expandDims, stack, gather, pad

ry:
cc @indutny

### DIFF
--- a/src/dataset.ts
+++ b/src/dataset.ts
@@ -156,11 +156,6 @@ class SliceDataset extends Dataset {
   }
 }
 
-// TODO
-function stack(tensors: Tensor[], axis = 0): Tensor {
-  throw Error("not implemented");
-}
-
 class BatchDataset extends Dataset {
   constructor(parent: Dataset, readonly batchSize: number) {
     super(parent);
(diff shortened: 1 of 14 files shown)


tools/dev_website.js
```diff
@@ -43,6 +43,7 @@ async function bundler(build) {
     outDir: wdir,
     production: prodFlag,
     publicUrl: "/",
+    souceMaps: false,
```
piscisaureus: did you experience any placebo effect?

src/dataset.ts
```diff
@@ -187,7 +182,8 @@ class BatchDataset extends Dataset {
       const out: NamedTensors = {};
       for (const name of Object.keys(batchComponents[0])) {
         const batch = batchComponents.map(tensors => tensors[name]);
-        out[name] = stack(batch, 0);
+        const first = batch.shift();
+        out[name] = first.stack(0, batch);
```
piscisaureus: I don't understand this (not what the deleted lines did). What does this do?

piscisaureus: Also I think there is a stack function somewhere that can now be removed?

tools/dev_website.js
ry: Yes. Removed this commit.

src/dataset.ts
ry: There's a bit in plot() that could be replaced with stack, but since that code will be deleted soon, I won't invest time in it. There was another tentative usage of stack in dataset - that was changed to use this new function.

PR #395, opened 2018-03-19
deps: add 'vega-embed' npm package

piscisaureus:


piscisaureus:
This was for CI only - continues in https://github.com/propelml/propel/pull/396

PR #397, opened 2018-03-20
notebook: various output handler improvements

piscisaureus:
* Use one output handler per cell, not one per output.
* Make the output handler responsible creating DOM text nodes.
* Rename 'console' RPC call to 'print()'.


piscisaureus:
cc @qti3e

PR #403, opened 2018-03-20
Use js example.

ry:


piscisaureus:
lgtm, but fix the tests

ry:
@piscisaureus PTAL

PR #404, opened 2018-03-20
Fix reduceMean test

piscisaureus:
@ritch does this patch make the api tests pass for you on gpu?


piscisaureus:
... and for you indutny?

piscisaureus:
... and for you @indutny?

PR #406, opened 2018-03-21
Implement reduceMin

qti3e:
and add tensor.size
I will need these APIs to implement NumPy's `array2string` (`tensor.toString()`)

https://github.com/numpy/numpy/blob/3629e3986333ecdf394924e53265e0d1061e928d/numpy/core/arrayprint.py#L1104-L1106

### DIFF
--- a/src/tensor.ts
+++ b/src/tensor.ts
@@ -505,6 +505,14 @@ export class Tensor implements types.Storage {
     return ops.reduceMax(this, axes, keepDims);
   }
 
+  /** Take the minimum value over the given axes.
+   * @param axes defaults to all.
+   */
+  reduceMin(axes?: number[], keepDims = false): Tensor {
+    if (!axes) axes = rangeJS(this.rank);
+    return ops.reduceMin(this, axes, keepDims);
+  }
+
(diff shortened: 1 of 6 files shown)


src/tensor.ts
```diff
@@ -169,6 +169,20 @@ export class Tensor implements types.Storage {
     return this.storage.shape;
   }
 
+  /** Returns size of the tensor
```
ry: Returns the total number of elements

src/tensor.ts
```diff
@@ -169,6 +169,20 @@ export class Tensor implements types.Storage {
...
+   *
+   *    import { range } from "propel";
+   *    range(200).reshape([100, 2]).size
+   */
+  get size(): number {
+    const shape = this.storage.shape;
```
ry: return this.shape.reduce((a, b) => a * b)

src/tensor.ts
qti3e: done

src/tensor.ts
qti3e: done

Issue #297, opened 2018-02-28
website: on /notebook show a list of the user's own notebooks


ry:

Qti3e:
How about having a profile page for each user? it's more general...

ry:
Yes, would be nice too

alaq:
I would like to take this issue.

ry:
Fixed by #407

PR #407, opened 2018-03-21
profile pages

alaq:
- [x] Created separate profile page per user.
- [x] Added last three notebooks by logged in user to the top of `/notebook`
	- [x] If user has no notebook, message is displayed (maybe not display anything instead? no header?)
- [x] User's names are clickable and redirecting to user's profile 

No error/warning in `./tools/tslint.js`
This addresses #297

### DIFF
--- a/website/db.ts
+++ b/website/db.ts
@@ -26,6 +26,7 @@ export interface Database {
   updateDoc(nbId: string, doc: NotebookDoc): Promise<void>;
   clone(existingDoc: NotebookDoc): Promise<string>;
   create(): Promise<string>;
+  queryProfile(uid: string, limit: number): Promise<NbInfo[]>;
   queryLatest(): Promise<NbInfo[]>;
   signIn(): void;
   signOut(): void;
@@ -167,6 +168,19 @@ class DatabaseFB implements Database {
     return out.reverse();
   }
 
+  async queryProfile(uid: string, limit: number): Promise<NbInfo[]> {
+    lazyInit();
+    const query = nbCollection.where("owner.uid", "==", uid).limit(limit);
+    const snapshots = await query.get();
+    const out = [];
+    snapshots.forEach(snap => {
+      const nbId = snap.id;
+      const doc = snap.data();
+      out.push({ nbId, doc });
+    });
+    return out;
+  }
+
   signIn() {
     lazyInit();
     const provider = new firebase.auth.GithubAuthProvider();
@@ -222,6 +236,15 @@ export class DatabaseMock implements Database {
   async create(): Promise<string> {
     this.inc("create");
     return "createdNbId";
+  }
+
+  async queryProfile(uid: string, limit: number): Promise<NbInfo[]> {
+    this.inc("queryProfile");
+    if (uid === defaultOwner.uid) {
+      return [{ nbId: "default", doc: defaultDoc }];
+    } else {
+      return [];
+    }
   }
 
   async queryLatest(): Promise<NbInfo[]> {
--- a/website/main.scss
+++ b/website/main.scss
@@ -54,6 +54,7 @@ body {
 }
 
 /* Centers and gives max-width to index, references, notebook pages */
+.centered,
 .training-example-inner,
 .use-it-inner,
 .flex-row,
@@ -654,7 +655,6 @@ header {
   font-size: 1.2rem;
   letter-spacing: 0.5px;
   display: flex;
-  margin-top: 8px;
 
   img {
     margin-right: $size0;
@@ -665,12 +665,6 @@ header {
   max-width: 960px;
   margin: 0 auto;
   padding: 0 8px;
-
-  .blurb {
-    box-shadow: inset 0 -1px 0 $borderColor;
-    background-color: #fff;
-    margin-bottom: 16px;
-  }
 }
 
 .most-recent {
@@ -728,33 +722,18 @@ header {
         background-color: $cellBackgroundFocus;
       }
 
-      &:hover .displayName {
-        color: $linkColor;
-        text-decoration: underline;
-      }
-
       p {
         margin: 8px;
         color: $base03;
       }
 
       .code-snippit {
-        margin: 8px;
+        max-height: 180px;
+        margin: 8px 8px 0 8px;
         font-family: 'Roboto Mono', monospace;
-      }
-
-      .blurb {
-        background-color: #fff;
-        box-shadow: inset 0 1px 0 $borderColor;
-        font-size: 1.3rem;
-        line-height: 1.6rem;
-        letter-spacing: 0.5px;
-        display: flex;
-        margin-top: 8px;
-
-        img {
-          margin-right: $size0;
-        }
+        white-space: pre;
+        overflow: hidden;
+        text-overflow: ellipsis;
       }
     }
   }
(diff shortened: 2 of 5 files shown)


website/nb.ts
```diff
@@ -464,7 +480,8 @@ export class MostRecent extends Component<any, MostRecentState> {
     // This is to avoiding calling into firebase during static HTML generation.
     if (IS_WEB) {
       const latest = await db.active.queryLatest();
-      this.setState({latest});
+      const own = await db.active.queryProfile(this.props.uid);
```
ry: When loading "/notebook" I get a crash here because `this.props.uid` is undefined.

website/nb.ts
```diff
@@ -687,6 +751,19 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
   }
 }
 
+function notebookList(notebooks: db.NbInfo[]): JSX.Element[] {
```
ry: Nice

website/db.ts
```diff
@@ -167,6 +168,19 @@ class DatabaseFB implements Database {
     return out.reverse();
   }
 
+  async queryProfile(uid): Promise<NbInfo[]> {
```
ry: `uid: string`

website/db.ts
```diff
@@ -167,6 +168,19 @@ class DatabaseFB implements Database {
...
+    lazyInit();
+    const query = nbCollection.where("owner.uid", "==", uid);
+    const snapshots = await query.get();
+    const out = [];
+    snapshots.forEach(snap => {
+      const nbId = snap.id;
+      const doc = snap.data();
+      out.unshift({ nbId, doc });
+    });
+    return out.reverse();
```
ry: If you used `push()` instead of `unshift()`, you wouldn't have to call `reverse()`

website/nb.ts
```diff
@@ -476,31 +493,78 @@ export class MostRecent extends Component<any, MostRecentState> {
...
+export interface ProfileState {
+  latest: db.NbInfo[];
+  profile: string;
+}
+
+export class Profile extends Component<any, ProfileState> {
+  async componentWillMount() {
+    // Only query firebase when in the browser.
+    // This is to avoiding calling into firebase during static HTML generation.
+    if (IS_WEB) {
```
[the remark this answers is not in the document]
alaq: Ok, I went ahead and removed it from the other location in the file, if that's okay.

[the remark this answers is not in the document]
ry: Yep, thanks!

alaq:
I believe the changes have been addressed. Let me know.

website/main.scss
```diff
@@ -639,20 +659,20 @@ header {
...
-  background-color: #fff;
-  box-shadow: inset 0 1px 0 #d8d8d8;
-  font-size: 1.2rem;
-  letter-spacing: 0.5px;
-  display: flex;
-  margin-top: 8px;
+.notebook-container {
+  max-width: 960px;
+  margin: 0 auto;
+  padding: 0 8px;
```
ry: This seems like an unrelated change. In fact, if we left out all the css changes completely - would it look terrible? I'd rather land the functionality first and then clean up the style later.

website/nb.ts
```diff
@@ -478,31 +497,79 @@ export class MostRecent extends Component<any, MostRecentState> {
...
+}
+
+export interface ProfileProps {
+  uid: string;
+  userInfo?: db.UserInfo;
+}
+
+export interface ProfileState {
+  latest: db.NbInfo[];
+  profile: string;
```
ry: ProfileState.profile is unused.

website/nb.ts
```diff
@@ -478,31 +497,79 @@ export class MostRecent extends Component<any, MostRecentState> {
...
+        ),
+      ),
+      h("ol", null, ...notebookList(this.state.latest)),
+    );
+  }
+}
+
+export interface ProfileProps {
+  uid: string;
+  userInfo?: db.UserInfo;
```
ry: ProfileProps.userInfo is unused.

website/nb.ts
```diff
@@ -478,31 +497,79 @@ export class MostRecent extends Component<any, MostRecentState> {
...
+  async componentWillMount() {
+    const latest = await db.active.queryProfile(this.props.uid);
+    this.setState({ latest });
+  }
+
+  render() {
+    if (!this.state.latest) {
+      return h(Loading, null);
+    }
+    const doc = this.state.latest[0].doc;
```
ry: It crashes here if `this.state.latest` is empty. You can do something simple for now like this:

```
if (this.state.latest.length === 0) {
  return h("h1", null, "User has no notebooks");
}
```

website/nb.ts
```diff
@@ -431,10 +431,12 @@ export class FixedCell extends Component<FixedProps, CellState> {
 export interface NotebookRootProps {
   userInfo?: db.UserInfo;
   nbId?: string;
+  uid?: string;
```
ry: This is getting a bit confusing with all these ids. So "uid" is supposed to be the user id of the profile we want to display, while "userInfo" is info about the logged in user. I'd suggest changing "uid" to "profileUid". You can leave "userInfo" as is, but I will change it in the future to be "currentUserInfo".

website/nb.ts
```diff
@@ -474,26 +491,28 @@ export class NotebookRoot extends Component<NotebookRootProps,
...
+export class MostRecent extends Component<MostRecentProps, MostRecentState> {
   async componentWillMount() {
-    // Only query firebase when in the browser.
-    // This is to avoiding calling into firebase during static HTML generation.
-    if (IS_WEB) {
-      const latest = await db.active.queryLatest();
-      this.setState({latest});
-    }
+    const latest = await db.active.queryLatest();
+    const own = await db.active.queryProfile(this.props.uid);
```
ry: It's a bit unfortunate that we're doing two queries... Maybe make a comment
`// TODO potentially these two queries can be combined into one.`

website/nb.ts
```diff
@@ -504,31 +523,77 @@ export class MostRecent extends Component<any, MostRecentState> {
...
         ),
         h("div", {"class": "most-recent-header-cta"},
           h("button", { "class": "create-notebook",
                         "onClick": () => this.onCreate(),
           }, "+ New Notebook"),
         ),
       ),
-      h("ol", null, ...notebookList),
+      empty,
+      h("ol", null, ...notebookList(this.state.own.slice(0, 3))),
```
ry: If you're only using 3 of the results, maybe limit the query to only return three?

alaq:
Requested changes have been applied, and @ry refactoring as well.
I believe this branch is ready to be merged.

Please review.

ry:
@piscisaureus Please review

piscisaureus:
The code changes look good to me.
However I think this is not rendering as intended: https://snag.gy/MOyb4N.jpg

ry:
That’s as intended. I’ll let a designer make a second pass.

ry:
Thank you @alaq ! This is a massive improvement to the website!

ry:
Profile pages uploaded to the main site. I'm finally able to explore @qti3e's notebooks : ) 
http://propelml.org/notebook/?profile=nsDPpi960tUR7Y5pmbsMs0zzkN42

Again thanks for trudging thru this @alaq

alaq:
My pleasure @ry 
Now that I can see all of mine too, I feel like I should be able to delete the ones that I am ashamed of

ry:
@alaq i had the same thought https://github.com/propelml/propel/issues/460

PR #409, opened 2018-03-21
nb: Don't display delete button if only one cell.

### LINKED ISSUE #323: Notebook allows last cell to delete
When inside a notebook, the last cell gives a user the option to delete the cell.
screen shot 2018-03-05 at 3 44 19 pm
When the user clicks delete, the notebook goes into an error state that cannot be recovered without refreshing:
screen shot 2018-03-05 at 3 44 26 pm
Expected experience is to not have the delete option on the last cell in a notebook.

ry:
Fixes #323.

Also improve MockDB and title editing.

(Part 1 of 2 in the nb cell refactor)

### DIFF
--- a/website/db.ts
+++ b/website/db.ts
@@ -16,6 +16,7 @@
 // This file contains routines for accessing the firebase database (firestore).
 // This is used to save and restore notebooks.
 // These routines are run only on the browser.
+import { assert } from "../src/util";
 
 // tslint:disable:no-reference
 /// <reference path="firebase.d.ts" />
@@ -28,7 +29,6 @@ export interface Database {
   queryLatest(): Promise<NbInfo[]>;
   signIn(): void;
   signOut(): void;
-  ownsDoc(doc: NotebookDoc): boolean;
   subscribeAuthChange(cb: (user: UserInfo) => void): UnsubscribeCb;
 }
 
@@ -96,7 +96,7 @@ class DatabaseFB implements Database {
   // Caller must catch errors.
   async updateDoc(nbId: string, doc: NotebookDoc): Promise<void> {
     if (nbId === "default") return; // Don't save the default doc.
-    if (!this.ownsDoc(doc)) return;
+    if (!ownsDoc(auth.currentUser, doc)) return;
     const docRef = nbCollection.doc(nbId);
     await docRef.update({
       cells: doc.cells,
@@ -178,11 +178,6 @@ class DatabaseFB implements Database {
     auth.signOut();
   }
 
-  ownsDoc(d: NotebookDoc): boolean {
-    const u = auth.currentUser;
-    return u && d && u.uid === d.owner.uid;
-  }
-
   subscribeAuthChange(cb: (user: UserInfo) => void): UnsubscribeCb {
     lazyInit();
     return auth.onAuthStateChanged(cb);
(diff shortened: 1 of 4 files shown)


website/db.ts
```diff
@@ -225,30 +231,33 @@ export class DatabaseMock implements Database {
...
+    for (const cb of this.authChangeCallbacks) {
+      cb(this.currentUser);
+    }
   }
 
   subscribeAuthChange(cb: (user: UserInfo) => void): UnsubscribeCb {
     this.inc("subscribeAuthChange");
-    return null;
+    this.authChangeCallbacks.push(cb);
+    return () => {
```
piscisaureus: What's this return value used for? It doesn't seem to do what the name of the return type suggests (rather than unsubscribe one listener, it clears out all the listeners)

website/nb_test.ts
```diff
@@ -23,128 +17,147 @@ testBrowser(function notebook_NotebookRoot() {
...
+  const deleteButton = document.querySelector(".delete-button");
+  assert(deleteButton == null);
+});
 
-  title = document.querySelectorAll("div.title > h2");
-  assert(1 === title.length);
-  assert("New Title" === title[0].innerHTML);
-  */
+// Call this to ensure that the DOM has been updated after events.
+function flush() {
```
piscisaureus: Can you give this function a return type?

website/nb_test.ts
```diff
@@ -23,128 +17,147 @@ testBrowser(function notebook_NotebookRoot() {
...
+  rerender();
```
piscisaureus: Is rerender() synchronous?

website/nb.ts
```diff
@@ -637,17 +648,17 @@ export class Notebook extends Component<NotebookProps, NotebookState> {
 
       const titleEdit = h("div", { class: "title" },
         h("input", {
-          class: "title-input",
-          onChange: event => this.onTypedTitle(event),
-          value: doc.title
+          "class": "title-input",
```
piscisaureus: Why did you add quotes?

website/nb.ts
```diff
@@ -225,10 +225,21 @@ export class Cell extends Component<CellProps, CellState> {
...
+  // component if the props have changed (e.g. if props.onDelete has been
+  // changed as demoed in the notebook_DeleteLastCell test).
   // Code updates are done in componentWillReceiveProps.
-  shouldComponentUpdate() {
-    return false;
+  // TODO This is very hacky. Ideally we wouldn't have to mange CM's
+  // state like this.
+  shouldComponentUpdate(nextProps, nextState) {
+    const propChanged = (name) =>
+      (this.props[name] == null && nextProps[name] != null) ||
```
piscisaureus: what about `(this.props[name] == null) !== (nextProps[name] == null)`

website/db.ts
```diff
@@ -225,30 +231,33 @@ export class DatabaseMock implements Database {
...
+      this.authChangeCallbacks = [];
+    };
   }
 }
 
-// Default to a mock, so none of the functions errors out during operation in
```
piscisaureus: Happy to see this gone!

website/nb.ts
piscisaureus: Nonetheless class is perfectly valid as an object key. And "value" and "onChange" are definitely not keywords.

ry: Fixed.

piscisaureus:
Lgtm after squashing

PR #410, opened 2018-03-22
Some changes to how progress bars are handled

piscisaureus:
* Use byte length, not string length, when computing the number of bytes
  loaded so far.
* handle http.get() errors.
* Reduce the number of type assertions.
* Explicitly indicate the end of a download by calling
  downloadProgress(job, null, null).

### DIFF
--- a/src/output_handler.ts
+++ b/src/output_handler.ts
@@ -37,7 +37,7 @@ const progressOutputHandlerMap = new Map<string, OutputHandler>();
 export class OutputHandlerDOM implements OutputHandler {
   // TODO colors should match those used by the syntax highlighting.
   private color = d3.scaleOrdinal(d3.schemeCategory10);
-  private progresses = new Map<string, Progress>();
+  private progressJobs = new Map<string, Progress>();
 
   constructor(private element: Element) {}
 
@@ -180,16 +180,16 @@ export class OutputHandlerDOM implements OutputHandler {
       // TODO: this isn't really correct - the progress bar might go backwards
       // when multiple parallel jobs are present and one completes before
       // the other.
-      this.progresses.delete(job);
+      this.progressJobs.delete(job);
       progressOutputHandlerMap.delete(job);
     } else {
-      this.progresses.set(job, progress);
+      this.progressJobs.set(job, progress);
     }
 
     const outputContainer = this.element.parentNode as HTMLElement;
     const progressBar = outputContainer.previousElementSibling as HTMLElement;
 
-    if (this.progresses.size === 0) {
+    if (this.progressJobs.size === 0) {
       // If there are no more downloads in progress, hide the progress bar.
       progressBar.style.width = "0"; // Prevent it from going backwards.
       progressBar.style.display = "none";
@@ -198,7 +198,7 @@ export class OutputHandlerDOM implements OutputHandler {
 
     let sumLoaded = 0;
     let sumTotal = 0;
-    for (const [_, {loaded, total}] of this.progresses) {
+    for (const [_, {loaded, total}] of this.progressJobs) {
       // Total may be null if the size of the download isn't known yet.
       if (total === null) {
         continue;
--- a/src/output_handler.ts
+++ b/src/output_handler.ts
@@ -16,9 +16,14 @@
 
 import * as d3 from "d3";
 import { createCanvas, Image } from "./im";
-import { Progress } from "./types";
 
 export type PlotData = Array<Array<{ x: number, y: number }>>;
+
+export interface Progress {
+  job: string;
+  loaded: number | null;
+  total: number | null;
+}
 
 export interface OutputHandler {
   imshow(image: Image): void;
@@ -27,11 +32,12 @@ export interface OutputHandler {
   downloadProgress(progress: Progress);
 }
 
-const progresses = {};
+const progressOutputHandlerMap = new Map<string, OutputHandler>();
 
 export class OutputHandlerDOM implements OutputHandler {
   // TODO colors should match those used by the syntax highlighting.
   private color = d3.scaleOrdinal(d3.schemeCategory10);
+  private progresses = new Map<string, Progress>();
 
   constructor(private element: Element) {}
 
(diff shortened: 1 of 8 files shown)


ry:
I haven't tested it but LGTM
@qti3e - any comments?

src/util.ts
```diff
@@ -196,67 +199,95 @@ async function fetch2(p: string, encoding: FetchEncoding)
...
+export function downloadProgress(job: string, loaded: number | null,
+                                 total: number | null): void {
+  if (activeOutputHandler) {
+    activeOutputHandler.downloadProgress({ job, loaded, total });
+    return;
+  }
+
+  if (IS_NODE) {
+    if (loaded === null && total === null) {
+      process.stdout.write("     \r");
```
qti3e: i guess it should be 6 spaces instead of 5 to remove `%`, ("99.99%".length === 6)

piscisaureus: Good catch! I'll add two spaces even as we might even have to unwrite "100.00%"

piscisaureus:
This PR isn't actually done. The DOM outputhandler needs to be changed to handle `downloadProgress(job, null, null).`

ry:
It would be very nice to have some sort of tests of this progress bar stuff...

src/util.ts
piscisaureus: fixed

src/util.ts
```diff
@@ -196,67 +199,96 @@ async function fetch2(p: string, encoding: FetchEncoding)
...
+  const http = nodeRequire(protocol === "https:" ? "https" : "http");
+  const job = randomString();
+
+  downloadProgress(job, 0, null); // Start download job with unknown size.
+
+  const chunks: Buffer[] = [];
+
+  try {
+    await new Promise((res, rej) => {
+      http.get(url, res => {
```
ry: You shadow res here

piscisaureus: Good point, will fix

src/util.ts
```diff
@@ -196,67 +199,96 @@ async function fetch2(p: string, encoding: FetchEncoding)
...
+
+  } finally {
+    downloadProgress(job, null, null); // End download job.
+  }
+
+  const buffer = Buffer.concat(chunks);
+  if (encoding === "utf8") {
+    return buffer.toString("utf8");
+  } else {
+    return buffer;
```
ry: I think you've already fixed this - but there was an error when fetchRemoteFile returned an ArrayBuffer.

https://gist.github.com/ry/852afb013d4b94f21eda26eb0e3e7fc4

piscisaureus: A Buffer is a subclass of ArrayBuffer so I thought it'd be okay to just return a buffer.

website/nb_test.ts
```diff
@@ -128,6 +128,67 @@ testBrowser(async function notebook_deleteLastCell() {
...
+  const visible = () => progressBar.style.display === "block";
+
+  // Call util.downloadProgress in the notebook sandbox.
+  const downloadProgress = async(job, loaded, total, cellId = 1) => {
+    const sandbox = nb.sandbox();
+    await sandbox.call(
+      "runCell",
+      `import { downloadProgress } from "test_internals";
+       downloadProgress(...${JSON.stringify([job, loaded, total])});`,
+      cellId);
```
ry: Yikes. I can see why this took a minute

website/nb_test.ts
```diff
@@ -128,6 +128,67 @@ testBrowser(async function notebook_deleteLastCell() {
...
+  await downloadProgress("job1", 0, null);
+  assert(visible());
+  assert(percent() === 0);
+  // Start another, size 30k bytes.
+  await downloadProgress("job2", 0, 30e3);
+  assert(visible());
+  assert(percent() === 0);
+  // Make progress on both jobs.
+  await downloadProgress("job1", 1e3, 10e3);
+  assert(percent() === 100 / 40);
```
ry: 100 / 40 ? I would have expected it to be a number between zero and one.

piscisaureus: it's an actual percentage

src/output_handler.ts
```diff
@@ -16,22 +16,28 @@
...
+
 export interface OutputHandler {
   imshow(image: Image): void;
   plot(data: PlotData): void;
   print(text: string): void;
   downloadProgress(progress: Progress);
 }
 
-const progresses = {};
+const progressOutputHandlerMap = new Map<string, OutputHandler>();
```
ry: `progressJobs` is less of a mouthful

src/output_handler.ts
```diff
@@ -16,22 +16,28 @@
...
 
 export class OutputHandlerDOM implements OutputHandler {
   // TODO colors should match those used by the syntax highlighting.
   private color = d3.scaleOrdinal(d3.schemeCategory10);
+  private progresses = new Map<string, Progress>();
```
ry: `progressJobs` is more descriptive

src/output_handler.ts
```diff
@@ -156,30 +162,54 @@ export class OutputHandlerDOM implements OutputHandler {
...
-      progressBar.style.display = "none";
-    } else {
-      progressBar.style.display = "block";
-      progressBar.style.width = (totalLoaded / totalTotal * 100) + "%";
-    }
+
+    // Avoid division by zero.
+    const percent = sumTotal > 0 ? sumLoaded / sumTotal * 100 : 0;
+    progressBar.style.display = "block";
+    progressBar.style.width = `${percent}%`;
```
ry: Is css ok with very precise floating point numbers? probably. otherwise round to integer.

src/util.ts
```diff
@@ -134,9 +133,13 @@ export function objectsEqual(a: any, b: any): boolean {
...
-export type FetchEncoding = "utf8" | "arraybuffer" | "buffer";
-// We return an ArrayBuffer on Web but a Buffer on Node.js
-export type BinaryData = Buffer | ArrayBuffer;
+
+export interface FetchEncodingMap {
+  "arraybuffer": ArrayBuffer;
+  "buffer": Buffer;
+  "utf8": string;
+}
+export type FetchEncoding = keyof FetchEncodingMap;
```
ry: This is fine for now - and an improvement on what was there before - but the real problem is the abstraction. fetch shouldn't handle encodings - it should handle data transfer. I will change this soon so that fetch only deals with ArrayBuffer objects.

src/util.ts
```diff
@@ -146,36 +149,36 @@ export type BinaryData = Buffer | ArrayBuffer;
...
-    }
-    req.open("GET", p, true);
-    req.responseType = encoding === "utf8" ? "text" : "arraybuffer";
-    if (encoding === "utf8") {
-      req.overrideMimeType("text/plain; charset=utf-8");
+    try {
+      const req = new XMLHttpRequest();
+      const onLoad = createResolvable();
+      req.onload = onLoad.resolve;
+      req.onprogress = ev => downloadProgress(job, ev.loaded, ev.total);
```
ry: very nice

src/util.ts
```diff
@@ -196,67 +199,97 @@ async function fetch2(p: string, encoding: FetchEncoding)
...
+    if (loaded === null && total === null) {
+      // Write 7 spaces, so we can cover "100.00%".
+      process.stdout.write(" ".repeat(7) + " \r");
+    } else if (!total) {
+      // Don't divide by zero.
+      return;
+    }
+
+    const now = Date.now();
+
```
ry: remove this blank line?

website/nb_test.ts
```diff
@@ -128,14 +128,87 @@ testBrowser(async function notebook_deleteLastCell() {
...
+  // Make progress on both jobs.
+  await downloadProgress("job1", 1e3, 10e3);
+  assert(percent() === 100 * 1e3 / 40e3);
+  await downloadProgress("job2", 1e3, 30e3);
+  assert(percent() === 100 * 2e3 / 40e3);
+  await downloadProgress("job2", 15e3, 30e3);
+  assert(percent() === 100 * 16e3 / 40e3);
+  // Set job1 to 100% from cellId 2.
+  await downloadProgress("job1", 10e3, 10e3, 2);
+  assert(percent() === 100 * 25e3 / 40e3);
```
ry: These will break if they're non integral - but adding assertClose seems annoying so it's fine.

ry:
@qti3e please review if you have the time

src/output_handler.ts
```diff
@@ -156,30 +162,54 @@ export class OutputHandlerDOM implements OutputHandler {
...
+    // TODO: this is really hacky and should not be the responsibility of the
+    // output handler at all.
+    const outputHandler = progressOutputHandlerMap.get(job);
+    if (outputHandler === undefined) {
+      progressOutputHandlerMap.set(job, this);
+    } else if (outputHandler !== this) {
+      return outputHandler.downloadProgress(progress);
+    }
+
+    if (progress.loaded === null) {
```
qti3e: To solve the todo, you can remove this if statement and use sth like:
```ts
this.progresses.set(job,{
  loaded: progress.loaded === null ? progress.total : progress.loaded,
  total: progress.total,
})
```
then later if `sumLoaded` is equal to `sumTotal` gc the `progressOutputHandlerMap` and `this.progresses`

src/output_handler.ts
piscisaureus: Seems to be no problem.

website/nb_test.ts
piscisaureus: I don't want to be overly pedantic here, but it *might* break if the value cannot be expressed as both a fraction of a (small) power of 2 and as a fraction of a power of 10.
25/40 can be written as 625/(10^3) and 5/(2^3), so it should be perfectly fine.

src/output_handler.ts
piscisaureus: Agree- fixed.

src/util.ts
piscisaureus: done

src/output_handler.ts
piscisaureus: Thanks for the suggestion, I think you're right, that might work.
I'm not going to do it however now -- you're welcome to send a followup PR.

PS - I hope you don't mind me rewriting large parts of your patch, I won't make a habit of it.
We really appreciate that you added the progress bar and the other contributions you've made. The issue here was that we were both working in the same part of the code base.

qti3e: @piscisaureus : )

PR #431, opened 2018-03-23
Add training example to homepage

ry:


ry:
@piscisaureus PTAL

src/experiment.ts
```diff
@@ -68,7 +68,14 @@ let lastPrint: Promise<void> = Promise.resolve();
 export async function print(...args: PrintArgs): Promise<void> {
   // All this is to make sure that log items are in the order they came.
   lastPrint = lastPrint.then(async() => {
-    console.log(await printHelper(...args));
+    // TODO this is a hack to get cell output working on the homepage. A better
+    // solution would be to globally replace the console object in sandbox.ts.
```
piscisaureus: I'm not entirely convinced that it would be better -- but that debate is for another day.

PR #436, opened 2018-03-23
imshow bug fixes

ry:
cc @qti3e

### DIFF
--- a/src/im.ts
+++ b/src/im.ts
@@ -44,19 +44,18 @@ function toTensor(data: Uint8Array, height: number, width: number,
 }
 
 export function toUint8Image(image: Tensor): Image {
-  const shape = image.shape;
   const dtype = image.dtype;
-  const height = shape[0] as number;
-  const width = shape[1] as number;
+  const height = image.shape[0] as number;
+  const width = image.shape[1] as number;
   let data;
-  if (shape.length === 2) {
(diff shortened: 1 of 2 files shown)


src/im.ts
```diff
@@ -44,19 +44,18 @@ function toTensor(data: Uint8Array, height: number, width: number,
...
+  if (image.shape.length === 2) {
     // convert to a 3D tensor
     image = image.reshape([height, width, 1]);
   }
-  if (shape.length === 3) {
-    const channels = shape[2];
+  if (image.shape.length === 3) {
+    const channels = image.shape[2];
     if (channels > 4 || channels === 2) {
-      throw new Error(`${shape} does not match any valid image mode.`);
```
piscisaureus: I don't think removing `new` is better.

ry: it's shorter

piscisaureus: There is this tslint rule that forces me to use `new Array` rather than `Array`. Why is this, and why does the same rule not apply to Error objects?

ry: yeah- but it's allowed. fine i'll change it back.

ry: Fixed.

piscisaureus: Agree I was being kinda pedantic.

piscisaureus:
LGTM

Issue #439, opened 2018-03-25
Should I await `experiment.sgd`?


indutny:
As title says, it is not clear whether user is supposed to await `sgd` result or not. The `sgd` returns promise, so I suppose - yes?

indutny:
Also, should it be `    await this.maybeSave();` in `experiment.ts`?

ry:
No - I've fixed the signatures in the above commit...

PR #440, opened 2018-03-26
sgd and minimize should return void.

ry:
To answer #439

### DIFF
--- a/src/experiment.ts
+++ b/src/experiment.ts
@@ -137,7 +137,7 @@ export abstract class Experiment {
   }
 
   /** Performs SGD given the loss and current parameters. */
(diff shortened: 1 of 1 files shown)


indutny:
What about `maybeSave` it seems to be doing something weird if not being awaited.

ry:
The idea was that it would save in the background to not block training. What is the weirdness?

indutny:
At times it saves several times in a row for consecutive sequence ids.

ry:
@indutny is that when you restart the process? Because it's told to save when the process exits, and when training starts. So restarting would cause that...

indutny:
No, it seems to just randomly happen from time to time during the learning, unless I've an await.

PR #441, opened 2018-03-26
Use browser like inspector to render output

qti3e:


website/serializer.ts
```diff
@@ -0,0 +1,252 @@
...
+function isNative(fn) {
+  return (/\{\s*\[native code\]\s*\}/).test("" + fn);
+}
+
+function Nodify(data): GNode<keyof NodeTypeMap> {
+  switch (typeof data){
+    case "number":
+      return {
+        type: "number",
+        data: data.toString()
```
piscisaureus: why is the number converted to the string here?

website/serializer.ts
```diff
@@ -0,0 +1,252 @@
...
+      };
+    case "string":
+      return {
+        type: "string",
+        data: formatString(data)
+      };
+    case "boolean":
+      return {
+        type: "boolean",
+        data: data ? 1 : 0
```
piscisaureus: why is a boolean converted to a number here?

website/serializer.ts
```diff
@@ -0,0 +1,252 @@
...
+          data: {
+            data: data.toString(),
+            dtype: data.dtype,
+            shape: data.shape
+          }
+        };
+      }
+      return {
+        type: "object",
+        cons: data.constructor.name
```
piscisaureus: Check that `data.constructor` is not null, because some objects may not have a constructor. E.g. those created with Object.create(null).

website/serializer.ts
```diff
@@ -0,0 +1,252 @@
...
+
+type Edge = [number, string, number];
+
+class Graph {
+  data: object = {};
+  edges: Edge[] = [];
+  id = 0;
+  map: WeakMap<object, number>;
+
+  constructor(data?, edges?) {
```
piscisaureus: I would suggest to simplify this to: `constructor(private data = {}, private edges: Edge[] = [])`.
Then you can remove the member variable declarations above and the assignments below.

website/serializer.ts
```diff
@@ -0,0 +1,252 @@
...
+    case "symbol":
+      return {
+        type: "symbol",
+        data: data.toString()
+      };
+  }
+  return undefined;
+}
+
+type Edge = [number, string, number];
```
piscisaureus: Can you add a comment explaining what these 2 numbers and 1 string represent.

website/serializer.ts
```diff
@@ -0,0 +1,252 @@
...
+    const node = Nodify(data);
+    let id;
+    if (node.type === "object" || node.type === "array") {
+      if (this.map.has(data)) {
+        return this.map.get(data);
+      }
+      id = this.push(node, parent, name);
+      this.map.set(data, id);
+      // insert childs
+      const keys: any[] = (Object.getOwnPropertyNames(data) as any[])
```
piscisaureus: The type to use here is `Array<string | symbol>`

website/serializer.ts
```diff
@@ -0,0 +1,252 @@
...
+  push(node: GNode<any>, parent?: number, name?: string) {
+    const nodeId = this.id++;
+    if (parent !== undefined) {
+      this.edges.push([parent, name.toString(), nodeId]);
+    }
+    this.data[nodeId] = node;
+    return nodeId;
+  }
+
+  get serialized() {
```
piscisaureus: Do not use a getter, just add a method called `serialize()`.
(getters are ok only if the function body is trivial, or there is another important reason to use them, e.g. compatibility with an existing API)

website/serializer.ts
```diff
@@ -0,0 +1,252 @@
...
+    return {data: this.data, edges: this.edges};
+  }
+
+  get object() {
```
piscisaureus: Same here - don't use a getter.

website/serializer.ts
```diff
@@ -0,0 +1,252 @@
...
+          space[key].data = node.data;
+        } else if (node.type === "object" || node.type === "array") {
+          space[key].data = {};
+          if (node.type === "object") {
+            space[key].cons = node.cons;
+          }
+        }
+      }
+    }
+    const edges = [...this.edges]; // clone
```
piscisaureus: Cloning this large array seem unnecessarily inefficient, you also shift the array on every loop iteration which is another copy, so you effectively have O(n*n) time complexity. Why don't you use a simple for loop?

website/serializer.ts
```diff
@@ -0,0 +1,252 @@
...
+        type: "symbol",
+        data: data.toString()
+      };
+  }
+  return undefined;
+}
+
+type Edge = [number, string, number];
+
+class Graph {
```
piscisaureus: Using this Graph class to deal with circular references works, but I think it may be a little over-engineered.
You could simply do this is the `nodify` function by adding an extra type, "backref". You'd do something along the lines:

```ts
const seenBefore = new Map<Object, number>();
let idCounter = 0;

function Nodify(data) {
  const id = idCounter++;
  if (data instanceof Object and seenBefore.has(data)) {
    return { type: "ref", id: seenBefore.get(data); }
  } else {
    seenBefore.set(data, id);
  }
  ... do the normal serialization steps here ...
}

**Comment by qti3e** [comment]
@piscisaureus PTAL : )

**Comment by piscisaureus** [comment]
Looks polish now: I ran into a couple of bugs:

`console.log(0)` <-- no output
`a={b:[0]}` <-- the second looks expanded but isn't. It'd be good if it were expanded by default, because this is a small object.
`console.log("a\\b\ncd")` <-- if you surround it with "" I think it should also escape special characters. Maybe use JSON.stringify(s) to get a quoted string.
`console.log([])` <-- weird ouput
`console.log({})` <-- looks like it can be expanded but there's nothing

**Comment by qti3e** [comment]
@piscisaureus @ry PTAL

File: src/util.ts [hunk 11]
```diff
@@ -130,7 +130,9 @@ export function objectsEqual(a: any, b: any): boolean {
   if (aProps.length !== bProps.length) return false;
   for (let i = 0; i < aProps.length; i++) {
     const k = aProps[i];
-    if (a[k] !== b[k]) {
+    if (typeof a[k] === "object") {
+      if (!objectsEqual(a[k], b[k])) return false;
+    } else if (a[k] !== b[k]) {
```
ry: This should probably be done as a separate commit? But I think #457 will take care of this.

website/inspector.tsx
```diff
@@ -0,0 +1,268 @@
...
+          }
+        }
+        break;
+      case "tensor":
+        properties.push({
+          object: {
+            data: data.data,
+            type: "tensorValue"
+          }
+        });
```
ry: missing break;

website/main.scss
```diff
@@ -281,6 +281,56 @@ header {
...
 }
 
+.tree-node {
+  .cm-s-syntax {
+    background-color: transparent;
+  }
+
+  pre {
+    background-color: transparent;
+    margin-top: -25px;
```
ry: Why this negative margin? Random margin values like this make it difficult to understand the layout. I'd prefer no margin.

src/util.ts
qti3e: @ry I'm waiting for #457, this one is not going to master anyway (it does not work with circular references)

qti3e:
```js
const largeArray = new Array(200).fill(2)
a = [2, 4, largeArray]
```
this is the only unexcpected result for now, but i had some difficulty fixing it (async behavior of preact)

website/inspector.tsx
qti3e: done

website/inspector_test.ts
```diff
@@ -0,0 +1,84 @@
...
+        },
+        "cons": "Object",
+        "isCircular": false
+      }
+    },
+    "cons": "Object"
+  });
+});
+
+test(async function inspector_serilizerRecursive() {
```
ry: serializerRecursive

website/inspector.tsx
```diff
@@ -0,0 +1,275 @@
...
+                { ...this.inlineRendered }
+              &#125;
+            </span>
+          );
+          break;
+        case "array":
+          value = (
+            <span>
+              <i class="cm-variable">
+                { inline ? "" : "Array " }
```
ry: Why use <i> here? It gives "Array" a different style than it is presented in the text. I think <span> is better here. (Also for Object)

website/inspector.tsx
```diff
@@ -0,0 +1,275 @@
...
+              </span>
+              &#123;
+                { ...this.inlineRendered }
+              &#125;
+            </span>
+          );
+          break;
+        case "array":
+          value = (
+            <span>
```
ry: unnecessary extra span

website/inspector.tsx
```diff
@@ -0,0 +1,275 @@
...
+    return false;
+  }
+
+  renderInlinePreview(data, type) {
+    let keys = Object.getOwnPropertyNames(data);
+    let elements = null;
+    if (type === "array") {
+      keys = keys.filter(x => x !== "length");
+    }
+    const maxElements = type === "array" ? 20 : 10;
```
ry: This is kinda odd logic. Let's leave it simply maxElements = 20 for now?

website/inspector.tsx
```diff
@@ -0,0 +1,275 @@
...
+          break;
+        case "undefined":
+          value = <span class="cm-atom">undefined</span>;
+          break;
+        case "null":
+          value = <span class="cm-atom">null</span>;
+          break;
+        case "object":
+          value = (
+            <span>
```
ry: unnecessary extra span?

website/inspector.tsx
qti3e: @ry it's necessary, only `Array` should have `.cm-variable` class.
Maybe use `div`?

website/serializer.ts
```diff
@@ -0,0 +1,237 @@
...
+export interface UnserializedNode extends GNode<any> {
+  isCircular: boolean;
+}
+export type UnserializedObject = ExtendableUnserializedObject |
+                                 UnserializedNode;
+
+// This class provides a graph data structure.
+// Which means it split objects into data and edges.
+// Data is an array (actually an object) of nodes. (look at GNode data-type)
+// We keep all references to children in one array called `edges`.
```
ry: I feel like this comment could give some context into how this adds to overall goals of Propel and the inspector...

website/inspector.tsx
qti3e: objects have keys, but arrays don't so to have the same amount of child in both cases we should half number of elements when it's object

website/inspector.tsx
ry: @qti3e div is `display: block`, so I guess just leave the span...But I don't understand why it's necessary TBH

qti3e: @ry 
because of this:
```js
[
  { ...this.inlineRendered }
]
```
jsx can not handle this:
```js
          value = (
              <span class="cm-variable">
                { inline ? "" : "Array " }
              </span>
              [
                { ...this.inlineRendered }
              ]
          );
```

website/inspector.tsx
ry: Maybe add a comment?

website/inspector.tsx
ry: Can you use an array of jsx elements instead ?

qti3e: yes but then I have to make `value` an array in all other places and use
```js
  { ...value }
```
instead of:
```js
  { value }
```
(line 269)
Do you want me to use an array?

piscisaureus:
LGTM, very nice.
There are still a few things that could be improved, but I think we'd better fix them incrementally after landing this patch.

PR #448, opened 2018-03-27
Fix npm package

ry:
Fixes a bunch of issues related to bundling.

### DIFF
--- a/src/disk_experiment_test.ts
+++ b/src/disk_experiment_test.ts
@@ -23,12 +23,12 @@ import { process } from "./util";
 import { isDir } from "./util_node";
 
 function setup() {
-  process.env.PROPEL_DIR = path.join(os.tmpdir(), "propel_test");
-  console.log("PROPEL_DIR", process.env.PROPEL_DIR);
-  if (fs.existsSync(process.env.PROPEL_DIR)) {
-    console.log("rm -rf PROPEL_DIR", process.env.PROPEL_DIR);
-    rimraf.sync(process.env.PROPEL_DIR);
-    fs.mkdirSync(process.env.PROPEL_DIR);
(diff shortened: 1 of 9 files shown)


src/util.ts
```diff
@@ -30,6 +30,9 @@ export const global = globalEval("this");
 export const IS_WEB = global.window !== undefined;
 export const IS_NODE = !IS_WEB;
 
+// Parcel trys to be smart and replaces the process object.
```
piscisaureus: tries

ry: Fixed.

src/disk_experiment_test.ts
```diff
@@ -19,23 +19,24 @@ import * as rimraf from "rimraf";
...
-  console.log("PROPEL_DIR", process.env.PROPEL_DIR);
-  if (fs.existsSync(process.env.PROPEL_DIR)) {
-    console.log("rm -rf PROPEL_DIR", process.env.PROPEL_DIR);
-    rimraf.sync(process.env.PROPEL_DIR);
-    fs.mkdirSync(process.env.PROPEL_DIR);
+  process.env.PROPEL_ROOT = path.join(os.tmpdir(), "propel_test");
+  console.log("PROPEL_ROOT", process.env.PROPEL_ROOT);
+  if (fs.existsSync(process.env.PROPEL_ROOT)) {
+    console.log("rm -rf PROPEL_ROOT", process.env.PROPEL_ROOT);
+    rimraf.sync(process.env.PROPEL_ROOT);
```
piscisaureus: { unlinkSync: fs.unlinkSync } ?

ry: It's only ever run outside the bundle, so it's not a problem.

It's not so unmanageable if we had tests on the bundle instead of pre-bundle - which is a necessity it seems.

Issue #423, opened 2018-03-22
Fix assertEqual


ry:
should be share code with `toEqual` in the jasmine_shim: https://github.com/propelml/propel/blob/master/tools/jasmine_shim.ts#L61-L84

should not depend on Tensor, so it can be used in util.ts 

however the assertEqual in tensor_util should continue to work with Tensors and also share code
https://github.com/propelml/propel/blob/master/src/tensor_util.ts#L71-L76

PR #457, opened 2018-03-29
Add assertEqual

dangerdak:
I attempted to address #423 , adding assertEqual to util.ts and sharing its code with tensor_util.ts and jasmine_shim.ts
Any guidance or feedback would be great @ry @piscisaureus

### DIFF
--- a/src/dataset_test.ts
+++ b/src/dataset_test.ts
@@ -15,8 +15,13 @@
 import { test } from "../tools/tester";
 import * as pr from "./api";
 import * as dataset from "./dataset";
-import { assert, assertAllClose, assertAllEqual, assertShapesEqual }
-  from "./tensor_util";
+import {
+  assert,
+  assertAllClose,
+  assertAllEqual,
+  assertEqual,
+  assertShapesEqual
+} from "./tensor_util";
 import { IS_NODE } from "./util";
 
 test(async function dataset_datasetFromSlices() {
@@ -142,7 +147,7 @@ test(async function dataset_iterableEndCondition() {
     const { features } = await p;
     count += features.shape[0];
   }
-  assert(count === 12);
+  assertEqual(count, 12);
 });
 
 test(async function dataset_shuffleSmoke() {
--- a/src/util.ts
+++ b/src/util.ts
@@ -76,6 +76,9 @@ export function equal(c: any, d: any): boolean {
       if (seen.get(a) === b) {
         return true;
       }
+      if (Object.keys(a).length !== Object.keys(b).length) {
+        return false;
+      }
       for (const key in { ...a, ...b }) {
         if (!compare(a[key], b[key])) {
           return false;
@@ -157,26 +160,6 @@ export function createResolvable<T>(): Resolvable<T> {
     methods = { resolve, reject };
   });
   return Object.assign(promise, methods) as Resolvable<T>;
-}
-
-export function objectsEqual(a: any, b: any): boolean {
-  const aProps = Object.getOwnPropertyNames(a);
-  const bProps = Object.getOwnPropertyNames(b);
-  if (aProps.length !== bProps.length) return false;
-  for (let i = 0; i < aProps.length; i++) {
-    const k = aProps[i];
-    if (a[k] !== b[k]) {
-      return false;
-    }
-  }
-  return true;
-}
-
-export function assertObjectsEqual(a, b) {
-  if (!objectsEqual(a, b)) {
-    console.error("Objects not equal:\n", a, "\n", b);
-    throw Error("assertObjectsEqual failed.");
-  }
 }
 
 let activeOutputHandler: OutputHandler | null = null;
--- a/src/util.ts
+++ b/src/util.ts
@@ -62,6 +62,41 @@ export function assert(expr: boolean, msg = "") {
   }
 }
 
+export function equal(c: any, d: any): boolean {
+  const seen = new Map();
+  return (function compare(a, b) {
+    if (a === b) {
+      return true;
+    }
+    if (typeof a === "number" && typeof b === "number" &&
+        isNaN(a) && isNaN(b)) {
+      return true;
+    }
+    if (a && typeof a === "object" && b && typeof b === "object") {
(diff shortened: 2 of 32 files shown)


src/dataset_test.ts
```diff
@@ -15,7 +15,13 @@
...
 import * as dataset from "./dataset";
-import { assert, assertAllClose, assertAllEqual, assertShapesEqual }
+import {
+  assert,
+  assertAllClose,
+  assertAllEqual,
+  assertEqual,
+  assertShapesEqual
+}
   from "./tensor_util";
```
ry: Style nit: Please have the closing curly brace and `from "./tensor_util";` on the same line.

src/dl/util.ts
```diff
@@ -17,6 +17,7 @@ import { DataType, DataTypeMap, NDArray } from "./math/ndarray";
  * =============================================================================
  */
 
+export { assertEqual } from "../util";
```
ry: Although this might have been violated elsewhere, the intention is that `src/dl` is standalone and imports should not reach outside of it. Please remove.

tools/jasmine_shim.ts
```diff
@@ -59,29 +59,12 @@ export const matchTesters: MatchTesters = {
...
-            return false;
-          }
-        }
-        seen.set(a, b);
-        return true;
-      }
+    try {
+      assertEqual(value, expected);
+      return true;
+    } catch (e) {
```
ry: Why is the try catch needed?

src/util.ts
```diff
@@ -62,6 +62,39 @@ export function assert(expr: boolean, msg = "") {
...
+  return (function compare(a, b) {
+    if (a === b) {
+      return true;
+    }
+    if (typeof a === "number" && typeof b === "number" &&
+        isNaN(a) && isNaN(b)) {
+      return true;
+    }
+    if (a && typeof a === "object" && b && typeof b === "object") {
+      if (seen.get(a) === b) {
```
ry: In this file there is a method called `objectsEqual` - can that be reused here?

src/util.ts
```diff
@@ -62,6 +62,39 @@ export function assert(expr: boolean, msg = "") {
...
+      return true;
+    }
+    return false;
+  })(c, d);
+}
+
+export function assertEqual(actual: any, expected: any, msg = null) {
+  if (!msg) { msg = `actual: ${actual} expected: ${expected}`; }
+  if (!equal(actual, expected)) {
+    throw new Error(msg);
```
ry: For debugging on web, in the case where the actual and expected are objects - it would be helpful to have an additional output here. Something like this:
```
console.error("assertEqual failed. actual = ", actual, "expected =", expected)
```
(This way the nice inspector interface can be used to inspect deep objects)

tools/jasmine_shim.ts
dangerdak: I put it there because assertEqual throws if its arguments aren't equal, whereas toEqual returns a boolean

ry: Ah makes sense - In that case can you add another routine to util.ts called `equal` which returns boolean, and use that here and in `assertEqual`, so the try block can be avoided?

src/util.ts
dangerdak: Great, thanks. Am I right in thinking that assertObjectsEqual (https://github.com/propelml/propel/blob/master/src/util.ts#L140) isn't needed anymore as assertEqual deals with objects?

ry: Yes, sir. Please replace assertObjectsEqual calls with asserEqual

src/dl/util.ts
```diff
@@ -73,6 +73,52 @@ export function assert(expr: boolean, msg: string) {
...
+    }
+    if (a && typeof a === "object" && b && typeof b === "object") {
+      if (seen.get(a) === b) {
+        return true;
+      }
+      if (!objectsEqual(a, b)) {
+        return false;
+      }
+      seen.set(a, b);
+      return true;
```
ry: It's hard for me to tell if this algorithm is correct. Could you add some tests to `util_test.ts` ?
(You can copy some from tools/jasmine_shim_test)

src/dl/util.ts
```diff
@@ -73,6 +73,52 @@ export function assert(expr: boolean, msg: string) {
...
+export function assertEqual(actual: any, expected: any, msg = null) {
+  if (!msg) { msg = `actual: ${actual} expected: ${expected}`; }
+  if (!equal(actual, expected)) {
+    console.error(
+      "assertEqual failed. actual = ", actual, "expected =", expected);
+    throw new Error(msg);
+  }
+}
+
+function equal(c: any, d: any): boolean {
```
ry: This should be in `src/util.ts` not `src/dl/util.ts`

src/dl/util.ts
```diff
@@ -73,6 +73,52 @@ export function assert(expr: boolean, msg: string) {
   }
 }
 
+export function assertEqual(actual: any, expected: any, msg = null) {
```
ry: This should be in src/util.ts not src/dl/util.ts

src/dl/math/ndarray.ts
```diff
@@ -222,7 +222,7 @@ export class NDArray<D extends DataType = DataType, R extends Rank = Rank> {
 
   asScalar(): Scalar<D> {
     this.throwIfDisposed();
-    util.assert(this.size === 1, "The array must have only 1 element.");
+    util.assertEqual(this.size, 1, "The array must have only 1 element.");
```
ry: Files in src/dl should not be changed.

src/dl/util.ts
dangerdak: Sure no problem. Just to check first, is this already covered by the tests in `jasmine_shim_test.ts`? As` toEqual` is tested in ` jasmine_shim.ts` and it just returns the value from this function. Shall i move the tests from `jasmine_shim_test.ts` to `util_test.ts` or copy them?
Thanks for your help, hope you don't mind all the questions!

src/util.ts
ry: great - thank you!

PR #465, opened 2018-03-29
Refactor internal fetch.

ry:


src/fetch.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+}
+
+// Transforms URLs that point to local resources to paths
+// matching those resources.
+export function resolve(p: string, isTest?: boolean): URL {
+  // If the optional isTest argument isn't supplied look for the global
+  // variable PROPEL_TESTER.
+  if (isTest == null) {
+    isTest = (global.PROPEL_TESTER != null && !!global.PROPEL_TESTER);
```
piscisaureus: Redundant, the part left of && can go.

src/fetch.ts
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  let base: URL;
+  if (IS_WEB) {
+    // Hack for tools/website_render.js
+    if (document.location.protocol === "about:") {
```
piscisaureus: What exactly causes this? is the url about:blank?

src/fetch.ts
```diff
@@ -0,0 +1,166 @@
...
+      base = new URL("http://localhost:8080/");
+    } else {
+      base = new URL(document.location);
+    }
+  } else {
+    base = new URL("file:///");
+    base.pathname = process.cwd();
```
piscisaureus: I know I suggested to do it this way, but in hindsight I think we need to call EncodeURIComponent. And decode it on the "way out", when we turn it back into an OS path.

src/fetch_test.ts
```diff
@@ -0,0 +1,66 @@
...
+      "../build/dev_website" + mnistPath);
+  }
+  assert(actual === expected);
+
+  // When not in testing mode, the URL shouldn't be modified.
+  actual = fetch.resolve(p, false).toString();
+  expected = p;
+  assert(actual === expected);
+
+  /*
```
piscisaureus: I'd like to see a test where you load a file with a space in the name.

src/im_test.ts
```diff
@@ -19,11 +19,12 @@ import { assertAllEqual } from "./tensor_util";
 import { assert, assertShapesEqual } from "./tensor_util";
 import { IS_NODE } from "./util";
 
+const url = "http://propelml.org/";
```
piscisaureus: Can we export this from a central place, maybe fetch.ts

src/fetch.ts
```diff
@@ -0,0 +1,172 @@
...
+  // fetchBrowserXHR, even tho it is actually sync.
+  const fs = nodeRequire("fs");
+  console.log("readFileSync", url.pathname);
+  const b = fs.readFileSync(url.pathname, null);
+  return b.buffer.slice(b.byteOffset, b.byteOffset + b.byteLength);
+}
+
+export function downloadProgress(job: string, loaded: number | null,
+                                 total: number | null): void {
+  if (activeOutputHandler) {
```
qti3e: I think you should use `getOutputHandler` instead

PR #477, opened 2018-03-31
[WIP] Add meshgrid

ritch:
Ignore whitespace changes...

### DIFF
--- a/src/api_test.ts
+++ b/src/api_test.ts
@@ -1311,3 +1311,12 @@ test(async function api_size() {
   assertEqual(api.range(20).size, 20);
   assertEqual(api.range(20).reshape([4, 5]).size, 20);
 });
+
+test(async function api_meshgrid() {
+  const x = api.linspace(1, 3);
+  const y = x.square();
+  // this is ugly... maybe just take TensorLike coordinates?
+  const coords = [await x.data(), await y.data()];
+  const [x1, y1] = api.meshgrid(coords);
+  console.log(x1.dataSync(), y1.dataSync());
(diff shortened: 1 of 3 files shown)


src/ops.ts
```diff
@@ -482,6 +482,21 @@ defBW("less", null, null); // Not differentiable.
 export let lessEqual = defFW("lessEqual", (x, y) => bo.lessEqual(x, y));
 defBW("lessEqual", null, null); // Not differentiable.
 
+export function meshgrid(inputs, indexing) {
```
ritch: Not sure if this should be an `op`...since it doesn't use `define*` or call other ops... Move the impl to `api.ts`?

ry: Having the implementation here is preferred even tho this is not a "backend op". See zeros as an example.

Rather than having `api.meshgrid` be a thin wrapper around this function, export ops.meshgrid from api directly. (Follow zeros example)

src/api_test.ts
```diff
@@ -1311,3 +1311,12 @@ test(async function api_size() {
   assertEqual(api.range(20).size, 20);
   assertEqual(api.range(20).reshape([4, 5]).size, 20);
 });
+
+test(async function api_meshgrid() {
+  const x = api.linspace(1, 3);
+  const y = x.square();
+  // this is ugly... maybe just take TensorLike coordinates?
```
ry: Yeah, that seems best.

ry:
Thanks - but closing this for now (clearing the queue)

Issue #236, opened 2018-02-08
Jupyter binds alt+enter to "run, then insert cell below". We should have the same


ry:

curt-mitch:
I could take this one if no one has started on it.

ry:
Please do!

Start here:
https://github.com/propelml/propel/blob/d56417c68edfba2848f0e76e9229e6ccd90ce7b0/website/nb.ts#L280-L289

And add tests following this example:
https://github.com/propelml/propel/blob/d56417c68edfba2848f0e76e9229e6ccd90ce7b0/website/nb_test.ts#L38-L60

Be warned that the notebook tests are a bit buggy (note the commented out part above) so you might encounter some difficulties.

curt-mitch:
Yeah, the feature implementation is straightforward but the test is proving difficult to implement. Could it be that the `nb.run` method needs to be run synchronously in some way? Getting to line 57 in `nb_test` reveals that the classList still contains both `notebook-cell-focus` and `notebook-cell-running`.

I'm also having trouble seeing whether `this.editor.getInputField().blur();` in line 286 of the snippet of nb.ts above actually triggers the "on blur" callback further below:
https://github.com/propelml/propel/blob/d56417c68edfba2848f0e76e9229e6ccd90ce7b0/website/nb.ts#L296-L298

curt-mitch:
@ry Just bumping this up. I could open a PR for the updated key bindings but not sure how much more productive I could be on fixing the flakey test.

ry:
@curt-mitch Yes, please do send the preliminary PR - I’ll see if I can figure out the test situation.

PR #480, opened 2018-04-03
For review: Key bindings for alt+enter to run then insert cell below

curt-mitch:
Addresses issue #236, but without added test per comment from @ry: https://github.com/propelml/propel/issues/236#issuecomment-377963926.


ry:
Yikes. I just realized this doesn't actually insert the cell below.  My bad. But merging it doesn't break anything - we just need to revisit this features along with the test soon.

PR #485, opened 2018-04-04
imshow: 4D images

qti3e:


src/im.ts
```diff
@@ -50,6 +50,10 @@ function toTensor(data: Uint8Array, height: number, width: number,
 }
 
 export function toUint8Image(image: Tensor): Image {
+  if (image.shape.length === 4) {
+    const [r, c, h, w] = image.shape;
```
ry: Shape should be n, h, w, c

src/im.ts
```diff
@@ -50,6 +50,10 @@ function toTensor(data: Uint8Array, height: number, width: number,
...
+    image = image.transpose([0, 2, 1, 3]).reshape([r * h, c * w]);
```
ry: It's not clear to me what's happening here. In the case we had a batch size 16, and 32 x 32 x 3 images - it would be nice to show those images in a 4x4 grid - that is a `4*32 x 4*32 x 3` image. The way I would do that is split the batch dim, so it has shape [4, 4, 32, 32, 3]
```
img = img.reshape([4, 4, -1])
```
Then transpose those batch dims to the height and width area, and reshape into a 3D tensor
```
img = img.transpose([2, 0, 3, 1, 4]).reshape([32 * 4, 32 * 4, 3])
```

Here's an example: http://propelml.org/notebook/?nbId=yQW73rabuUBeTa72OiHL

src/im.ts
qti3e: @ry based on your example I wrote this code and my expected usage was this:
(`c` stands for **column** here not to be confused with **channel**)
```ts
import * as pr from "propel";
let ds = pr.dataset("mnist/test").batch(15);
let { images } = await ds.next();
pr.imshow(images.reshape([3, 5, 28, 28]));
```
but it think you want sth like this one:
```ts
import * as pr from "propel";
let ds = pr.dataset("mnist/test").batch(15);
let { images } = await ds.next();
pr.imshow(images.reshape([15, 28, 28, 1]));
```
?

src/im.ts
```diff
@@ -221,6 +221,19 @@ export async function imread(filename: string, mode: Mode = "RGBA")
...
                              filename: string,
                              handler?: "PNG" | "JPEG"): Promise<void> {
+  // Force a valid file extension at the end of `filename` (default to .png)
+  // test: https://goo.gl/jhdWct
+  filename = String(filename)
+    .replace(/\..+[^(png|jpg|jpeg)]$|[^(png|jpg|jpeg)]$/i, "$&.png");
+  if (tensor.rank === 4) {
+    const [n, h, w, c] = tensor.shape;
+    for (let i = 0; i < n; ++i) {
+      // Regex test: https://goo.gl/22NXUp
```
ry: Could you export a global variable here with the regexp, and add a test to im_test demoing these instead of this link? If the RegExp is modified (EG by adding another image type) it can potentially break in bad ways now.

src/matplotlib.ts
```diff
@@ -71,6 +71,13 @@ export function imshow(tensor: Tensor): void {
     console.warn("imshow: no output handler");
     return;
   }
+  if (tensor.rank === 4) {
+    const [n, h, w, c] = tensor.shape;
+    for (let i = 0; i < n; ++i) {
+      imshow(tensor.slice(i, 1).reshape([h, w, c]));
```
ry: I think there should be a limit here on the number of images that can be displayed. Maybe if the combined width or height exceeds 2000px? It should do something like `console.log("Truncated imshow, due to size.")`

src/im.ts
```diff
@@ -221,6 +221,19 @@ export async function imread(filename: string, mode: Mode = "RGBA")
 export async function imsave(tensor: Tensor,
```
ry: Could you add a smoke test that does imsave and imshow on some images say, `zeros([4, 32, 32, 3])`?

qti3e: done

src/matplotlib.ts
qti3e: done

src/im.ts
qti3e: I wrote a separate function called `formatImageName`

src/im_test.ts
```diff
@@ -116,3 +118,15 @@ test(async function im_toUint8ImageGrayscale() {
...
+test(async function im_imsave() {
+  const image = zeros([4, 32, 32, 3]);
+  if (!IS_NODE) return;
+  const path = nodeRequire("path");
+  const tmpDir = tmpdir();
+  await imsave(image, path.join(tmpDir, "im.png"));
+  assert(isFile(path.join(tmpDir + "im-0.png")));
+  assert(isFile(path.join(tmpDir + "im-1.png")));
+  assert(isFile(path.join(tmpDir + "im-2.png")));
+  assert(isFile(path.join(tmpDir + "im-3.png")));
```
ry: `fs.existsSync()` should be sufficient here.

src/util_test.ts
```diff
@@ -95,3 +95,18 @@ test(async function util_captureStackTrace() {
   const s = util.captureStackTrace();
   assert(/util_captureStackTrace/.test(s));
 });
+
+test(async function util_formatImageName() {
```
ry: Nice!

src/im_test.ts
qti3e: it's deprecated

Issue #469, opened 2018-03-30
Params interface doesn't show up in docs.


ry:
The bug is somewhere in https://github.com/propelml/propel/blob/master/tools/gendoc.ts

dangerdak:
[params.ts](https://github.com/propelml/propel/blob/1e76e609ae099898c8e2573a2809927fb5175f62/src/params.ts#L33) exports an interface and [gendoc.ts doesn't visit interfaces](https://github.com/propelml/propel/blob/master/tools/gendoc.ts#L182).
I tried adding a `visitInterface` function (a slightly modified version of [visitClass](https://github.com/propelml/propel/blob/1e76e609ae099898c8e2573a2809927fb5175f62/tools/gendoc.ts#L268) - just so it accepts the right types) but then all of these interfaces show up in the docs which seems undesirable:
```
interface Params
interface ParamsFn
interface ExperimentOpts
interface Promise
interface SGDOpts
interface MinimizeResult
interface Array
interface Storage
interface FormatOptions
interface PoolOpts
interface LinearOpts
interface ConvOpts
interface BatchNormOpts
interface TensorOpts
interface ConvOpts
```

ry:
Cool - thanks for looking into this. Those are all public interfaces - I'd say let's start with having them all in the docs - and then later devise a way to minimize noise in the index...  I can think of a few strategies for that - one could be a special jsdoc comment that makes an interface not get displayed. Or, maybe it would be better to have a way to attach the documentation for an interface to a method. For example, in the case of PoolOpts or ConvOpts it would be ideal to list the structure inside the maxPool() or conv2d() methods respectively. In any case, it's more correct to display all the public interfaces than it is to not display them, so let's start with that.

ry:
Thanks @dangerdak! I enjoyed [your blog post](https://medium.com/yld-engineering-blog/machine-learning-in-javascript-with-propel-50bfa46e8b03) too!

dangerdak:
Thanks dude :)

PR #493, opened 2018-04-05
Show interfaces in docs

dangerdak:
Params wasn’t showing up as gendoc.ts wasnt visiting interfaces (#469 ).

To show the interfaces in the docs I added `visitClass` in the [`ts.isInterfaceDeclaration` check](https://github.com/propelml/propel/compare/master...dangerdak:params-docs?expand=1#diff-324e575d0900c95fbd2bbfc469054499R188) but to do this I had to change type signatures for `visitClass`, `classElementName` and `visitProp` - not sure if this is a bad way to do things :grimacing:
I also added `ts.isMethodSignature` and `ts.isPropertySignature` checks to the `visitClass` function.
... Maybe there’s some nicer way to organise all this?

This caused some interfaces from `deps/` to show up (eg `Array` and `Promise`) so I filtered out nodes which contain the excludes from tsconfig.json in their declarations file paths. 

ConvOpts appears duplicated in the docs as it’s defined in both [src/types.ts](https://github.com/propelml/propel/blob/3aae62dcffda91500574a2f5b201fab9d6a75124/src/types.ts#L27) and [src/layers.ts](https://github.com/propelml/propel/blob/f5e25608af06b4abdfcb45cd8f1a229511e2ee7b/src/layers.ts#L47)

And I’m not sure why the [`MinimizeResult` interface](https://github.com/propelml/propel/blob/41c53c27cec0f8b97d98a728e340e7dfcc8dee20/src/optimizers.ts#L36) is showing up in the docs despite not being exported - is this expected?

### DIFF
--- a/tools/gendoc.ts
+++ b/tools/gendoc.ts
@@ -82,6 +82,7 @@ function getGithubUrlForFile(fileName: string) {
 
 export function genJSON(): DocEntry[] {
   // Global variables.
+  const { exclude } = require("../tsconfig");
   const visitQueue: ts.Node[] = [];
   const visitHistory = new Map<ts.Symbol, boolean>();
   let checker: ts.TypeChecker = null;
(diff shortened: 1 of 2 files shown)


ry:
It looks good! Thank you!

Regarding ConvOpts - there's actually two versions of it - which is poor design on my part. I will rename one of soon.

Regarding MinimizeResult - because sgd() is exported and sgd() depends on MinimizeResult, it is part of the public interface.

Regarding test failure. It should be fixed after rebasing.

dangerdak:
Thanks for the explanations :)
I rebased but that just moved the test failure along - it does pass locally and on appveyor though so it seems a bit odd that just travis thinks the generated docs have length 0. I double checked that I've rebased the latest changes and tried using the same node version as travis, and still it passes locally... not sure what else to try tbh!

ry:
@dangerdak The problem is that travis builds in `/home/travis/build/propelml/propel/` and `build` is one of the excluded filenames : )

Here's a patch I was using to debug it: https://gist.github.com/ry/4e5dc194debf3ff6d286fb82394f1789

dangerdak:
aah, thank you :)

dangerdak:
@ry it's passing the tests at last :relieved:  - PTAL

PR #494, opened 2018-04-05
inspector: Call toggle correctly.

ry:
cc @qti3e


ry:
closed in favor of #497

PR #497, opened 2018-04-06
Inspector improvements

piscisaureus:


website/inspector.tsx
```diff
@@ -14,263 +14,183 @@
...
     }
-    const isCollapsedClass = isCollapsed ? " collapsed" : "";
-    const isCollapsibleClass = isCollapsible ? " collapsible" : "";
-    const inlineClass = inline ? " inline" : "";
-    const depthClass = " depth-" + depth;
-    const className = "tree-node" + isCollapsedClass + inlineClass + depthClass;
+
+    return [<a href="javascript:void(0)"
+               class="inspector-expand-button"
+               onClick={ this.toggle } />];
```
ry: need to .bind(this)

src/util.ts
```diff
@@ -179,6 +179,14 @@ export function randomString(): string {
   return (Math.random() + 1).toString(36).padEnd(12, "0").slice(2, 12);
 }
 
+// Returns true if a key is numerical and cam be used as an array or string
+// index.
+const maxIndex = 2 ** 32 - 1;
+export function isNumericalKey(key: string): boolean {
+  return +key <= maxIndex &&
```
ry: `Number(key)` instead of `+key`
And why does it need to be compared to maxIndex?

website/inspector.tsx
```diff
@@ -14,263 +14,176 @@
...
+        { ...this.renderExpandButton(this.props.object) }
+        { ...this.renderValue(this.props.object) }
       </div>
     );
   }
+
+  private toggle(event: MouseEvent) {
+    const el = event.currentTarget as HTMLElement;
+    const parent = el.parentNode as HTMLElement;
+    parent.classList.toggle("inspector-collapsed");
```
ry: This goes against react principles... the class should be set in the `render()` method.

website/inspector.tsx
```diff
@@ -14,263 +14,176 @@
...
-  }
-
-  shouldExpandByDefault({ depth = 0, object }: InspectorProps,
-                        isCollapsible: boolean): boolean {
-    if (!isCollapsible) return false;
-    if (object.isCircular) return false;
-    if (depth === 0) return true;
-    return Object.getOwnPropertyNames(object.data).length <= 15;
-  }
+export interface InspectorState {}
```
ry: Why remove `isCollapsed` from here?

website/main.scss
```diff
@@ -268,50 +268,73 @@ header {
...
+}
 
-  pre {
-    background-color: transparent;
+.inspector,
+.inspector-property {
+  width: 100%;
+}
+
+.inspector-expand-button {
```
ry: I think this should be called `.inspector__expand-button` following the BEM scheme: 
https://en.bem.info/methodology/quick-start/
https://github.com/propelml/propel/blob/master/website/main.scss#L1

website/main.scss
```diff
@@ -268,50 +268,73 @@ header {
...
+  &::after {
+    content: "◢";
   }
+}
 
-  .tree-arrow-wrapper {
-    display: inline-block;
-    width: $size1;
-    height: $size1;
+.inspector-property-list {
```
ry: `.inspector__property-list`

website/nb_transpiler.ts
```diff
@@ -491,6 +491,7 @@ class MappedString extends Array<MappedChar> {
   }
 
   toString(): string {
+    console.log(this.reduce((str, c) => str + c.char, ""));
```
ry: Remove debugging output

website/nb_transpiler.ts
```diff
@@ -100,7 +100,7 @@ const evalScopeVisitors = {
       return;
     }
 
-    state.edit.insertBefore(node, `__global.${node.id.name}=`);
+    state.edit.insertBefore(node, `void (__global.${node.id.name}=`);
```
ry: separate commit?

ry:
Inspection should be collapsed by default.

screen shot 2018-04-07 at 10 43 19 am

website/nb_transpiler.ts
piscisaureus: Moved to #501

src/util.ts
piscisaureus: > And why does it need to be compared to maxIndex?
Since the spec says that the the maximum array key is UINT32_MAX. The effect is this:

```
> a = []; a[10] = "test"; console.log(a.length) 
11
> a = []; a[9999999999] = "test"; console.log(a.length)
0
```

website/inspector.tsx
```diff
@@ -14,263 +14,194 @@
...
     }
-    const isCollapsedClass = isCollapsed ? " collapsed" : "";
-    const isCollapsibleClass = isCollapsible ? " collapsible" : "";
-    const inlineClass = inline ? " inline" : "";
-    const depthClass = " depth-" + depth;
-    const className = "tree-node" + isCollapsedClass + inlineClass + depthClass;
+
+    const id = randomString();
+    return [<a class="inspector__expand-button"
+               href={ `javascript:Inspector.toggle("${id}")` }
```
ry: Add some comments about why this is being done this way.

website/inspector.tsx
```diff
@@ -14,263 +14,194 @@
...
   }
+
+  static toggle(id: string): void {
+    const el = document.getElementById(id);
+    const parent = el.parentNode as HTMLElement;
+    parent.classList.toggle("inspector__collapsed");
+  }
+}
+
+if (IS_WEB) {
```
ry: We don't use jsdom anymore - so this will only be loaded on web. IS_WEB guards are not necessary.

website/inspector.tsx
```diff
@@ -14,263 +14,194 @@
...
+      elements = [...elements, ...this.renderValue(d.primitive)];
+    }
+    // Add the list of properties.
+    if (d.props.length > 0 || !d.primitive) {
+      // Add " " if necessary.
+      if (hasInterestingConstructor || d.primitive) {
+        elements.push(" ");
+      }
+      // Add { or [;
+      elements.push(isArray ? "[" : "{");
```
ry: When expanding an array, square brackets are used, but I think this is not correct as it is using labels...

screen shot 2018-04-09 at 9 33 00 am

website/inspector.tsx
```diff
@@ -14,263 +14,347 @@
...
-        <span class="cm-keyword">[Circular]</span>
-      );
+    const d = this.props.descriptors;
+    return <div class="inspector cm-s-syntax">{ this.renderRoot(d) }</div>;
+  }
+
+  static toggle(id: string): void {
+    const el = document.getElementById(id);
+    const parent = el.parentNode as HTMLElement;
+    parent.classList.toggle("inspector__li--expanded");
```
ry: Can you add a comment here about why react state isn't used - i.e. to simplify event handling after static rendering

website/main.scss
```diff
@@ -268,53 +268,122 @@ header {
...
-  &.inline {
-    display: inline-block;
-    text-align: center;
+  .inspector__content {
+    width: auto;
+    overflow: hidden;
+    white-space: nowrap;
+  }
+  .inspector__content--pre {
+    white-space: pre;
```
ry: this is cute

website/main.scss
```diff
@@ -268,53 +268,122 @@ header {
...
+  }
+
+  .inspector__list { 
+    padding: 0;
 
-    .tree-arrow-wrapper {
+    &,
+    li,
+    .inspector__content {
```
ry: To avoid "No Descending Specificity", do not nest these css classes. See https://github.com/propelml/propel/blob/adffe44b55aa14cfa577ce4aeabe8a4db2c10f6d/website/index.scss as an example

website/sandbox.ts
```diff
@@ -79,7 +79,7 @@ class Console {
   constructor(private rpc: RPC, private cellId: number) { }
 
   private print(...args: any[]) {
-    this.rpc.call("print", this.cellId, args.map(serialize));
+    this.rpc.call("print", this.cellId, args.map(v => describe(v)));
```
ry: `args.map(describe)` ?

ry:
Also please add an issue detailing the unit tests needed.

piscisaureus:
@ry PTAL

website/serializer.ts
```diff
@@ -13,225 +13,265 @@
...
+  type: "function";
+  name: string;
+  async: boolean;
+  class: boolean;
+  generator: boolean;
+}
+export interface ObjectDescriptor extends BaseObjectDescriptor {
+  type: "object";
+}
+export interface TensorDescriptor extends BaseObjectDescriptor{
```
ry: style

tslint doesn't have a rule for this https://github.com/palantir/tslint/issues/746 
However according to this comment: https://github.com/palantir/tslint/issues/746#issuecomment-311455830 it can be achieved with tslint-eslint-rules by adding an object-curly-spacing linting rule. I will experiment with it.

website/serializer.ts
```diff
@@ -13,225 +13,265 @@
...
+
+export interface AtomDescriptor {
+  type: "null" | "undefined" | "getter" | "gettersetter" | "setter";
+}
+export interface PrimitiveDescriptor {
+  type: "boolean" | "date" | "number" | "regexp" | "string" | "symbol";
+  value: string;
+}
+export interface BaseObjectDescriptor {
+  ctor: string | null;            // Name of the object's constructor.
```
ry: `ctor?: string`

website/serializer.ts
```diff
@@ -13,225 +13,265 @@
...
+      } as ArrayDescriptor;
+    } else if (value instanceof Tensor) {
+      // Tensor shapes [] and [1] are equivalent, but the latter looks better.
+      const tensor = value.shape.length > 0 ? value : value.reshape([1]);
+      d = {
+        type: "tensor",
+        dtype: tensor.dtype,
+        shape: [...tensor.shape],
+        formatted: formatTensor(tensor)
+      } as TensorDescriptor;
```
ry: Here `d` is missing `ctor` and
```
Property 'props' is missing in type '{ type: "tensor"; dtype: DType; shape: number[]; formatted: string; }'
``` 

I saw this by running `ts-node website/inspector_test.ts` ... not sure why tsc doesn't pick up this error?

website/inspector.tsx
```diff
@@ -14,263 +14,363 @@
...
-    const { object: { data, type }, depth = 0 } = this.props;
-    const properties = [];
-    switch (type) {
-      case "object":
+    // Render the list of properties and other content that is expanded when the
+    // expand button is clicked.
+    const listItems: ElementList = [];
+
+    if (d.type === "tensor") {
+        // TODO: the tensor is currently formatted by Tensor.toString(),
```
piscisaureus: Style

website/inspector.tsx
```diff
@@ -14,263 +14,363 @@
...
+        // Square brackets are used and they are always visible.
+        //   [1, 2, 3, 4, 5, foo: "bar"]
+        //   Array(3) [1, 2, 3, ...properties...]
+        //   Float32Array(30) [...]
+        const ctorVisible = d.ctor === "Array" ? "expanded" : "always";
+        elements.push(
+          ...showWhen(ctorVisible,
+                      <span class="cm-def">{ d.ctor }</span>,
+                      "(", <span class="cm-number">{ d.length }</span>, ") "),
+          "[", list, "]"        );
```
piscisaureus: Style

website/serializer.ts
```diff
@@ -13,225 +13,265 @@
...
 
-// object, property name, value
-export type Edge = [number, string, number];
-
-export interface SerializedObject {
-  data: { [key: number] : GNode<any> };
-  edges: Edge[];
+class Placeholder {
+  static getter = new Placeholder("getter");
+  static gettersetter = new Placeholder("getter");
```
piscisaureus: Typo: getter -> gettersetter

website/inspector_test.ts
```diff
@@ -1,84 +1,271 @@
...
 import { ones } from "../src/api";
-import { assert, assertEqual } from "../src/util";
+import { assertEqual } from "../src/util";
 import { test, testBrowser } from "../tools/tester";
 import { Inspector } from "./inspector";
-import { serialize, unserialize } from "./serializer";
+import { describe, DescriptorSet, InspectorOptions } from "./serializer";
 
-// tslint:disable:object-literal-sort-keys
+test(async function inspector_describe() {
```
ry: 👍

website/serializer.ts
```diff
@@ -13,225 +13,265 @@
...
+export interface AtomDescriptor {
+  type: "null" | "undefined" | "getter" | "gettersetter" | "setter";
+}
+export interface PrimitiveDescriptor {
+  type: "boolean" | "date" | "number" | "regexp" | "string" | "symbol";
+  value: string;
+}
+export interface BaseObjectDescriptor {
+  ctor: string | null;            // Name of the object's constructor.
+  props: PropertyDescriptor[];    // Descriptors for object properties.
```
ry: `props?: PropertyDescriptor[];`

Seems there are several descriptors that don't specify props.

website/main.scss
```diff
@@ -268,50 +268,91 @@ header {
...
+}
+
+.inspector__li {
+  display: flex;
+}
+
+.inspector__toggle {
+  display: inline-block;
+  margin: 0 0 0 -1.25em;
+  width: 1.25em;
```
ry: There are several random spacing values here `1.25em` and `1.75em`. It would be preferable to use `$size0` or `$size1`. Note that px values and em values (at least `rem` values) are equivalent due to this bit https://github.com/propelml/propel/blob/39f885b420709c5a7e447ca673fb48b8e1cf19ea/website/main.scss#L48-L52

website/sandbox.ts
```diff
@@ -78,20 +78,25 @@ function guessCellId(error?: Error): number {
 class Console {
   constructor(private rpc: RPC, private cellId: number) { }
 
-  private print(...args: any[]) {
-    this.rpc.call("print", this.cellId, args.map(serialize));
+  private print(data: InspectorData) {
+    this.rpc.call("print", this.cellId, data);
+  }
+
+  inspect(value: any, options?: InspectorOptions) {
```
ry: I'm concerned about adding methods to `Console` which are not part of the spec: https://developer.mozilla.org/en-US/docs/Web/API/Console

website/serializer.ts
piscisaureus: But all descriptors that are object descriptors do!

website/serializer.ts
piscisaureus: I'm fighting with the type system here... The descriptor at this point is partially conforming, `ctor` and `props` are added below.

website/serializer.ts
piscisaureus: It's not an optional field; the value is either a string or (in a very rare circumstance) explicitly null.

website/inspector.tsx
piscisaureus: There are two comments about this elsewhere, I think that's enough.

website/serializer.ts
piscisaureus: Fixed.

website/main.scss
piscisaureus: I want to keep it as-is if that's okay. I used "em" units because I thought the indentation should be related to font size.

website/sandbox.ts
piscisaureus: Renamed to console._inspect as per discussion offline.

website/sandbox.ts
piscisaureus: fixed

website/serializer.ts
piscisaureus: Fixed

piscisaureus:
@ry ptal

website/main.scss
ry: px and font size are already related due to that snippet above.

website/serializer.ts
ry: ah ok

website/serializer.ts
ry: ok

PR #508, opened 2018-04-11
Fix bug in layers.conv2d and add util.captureStackTrace for debugging ops.

ry:


src/util.ts
```diff
@@ -182,3 +182,18 @@ export function randomString(): string {
 export function tmpdir(): string {
   return process.env.TEMP || process.env.TMPDIR || "/tmp";
 }
+
+// TODO maybe share this with nb_transpiler.getEntryPoint().
+export function captureStackTrace(error?: Error): string {
```
piscisaureus: Why the optional arg?

ry: So it can potentially be used by nb_transpiler.getEntryPoint() 
https://github.com/propelml/propel/blob/39f885b420709c5a7e447ca673fb48b8e1cf19ea/website/nb_transpiler.ts#L378-L392

piscisaureus:
Lgtm

PR #513, opened 2018-04-15
Various clean ups - including defaulting to webgl on web 

ry:


src/dl/math/backends/webgl/select_gpu.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  constructor(shape: number[]) {
+    this.outputShape = shape.slice();
+    const dType = getCoordsDataType(shape.length);
+    const sampleCoords = getSampleCoords(shape.length);
+
+    this.userCode = `
+      void main() {
+        ${dType} coords = getOutputCoords();
+        int cond = int(getCond(${sampleCoords}));
```
piscisaureus: just for my understanding - where does getCond come from?

src/tensor_util.ts
```diff
@@ -211,3 +211,7 @@ export function inferShape(arr: number | boolean | RegularArray<number> |
   }
   return shape;
 }
+
+export function isTensor(t: any): t is Tensor {
+  return t.id && !!t.dtype && !!t.device;
```
piscisaureus: Why not (t instanceof Tensor) ?

src/tensor.ts
```diff
@@ -24,7 +24,14 @@ import { assert, assertEqual, IS_NODE } from "./util";
 
 export function convert(t: types.TensorLike,
                         opts?: types.TensorOpts): Tensor {
-  if (t instanceof Tensor) return t;
+  if (t instanceof Tensor) {
```
piscisaureus: Why do you use `isTensor` elsewhere, but `instanceof Tensor` here?

src/tensor.ts
```diff
@@ -22,9 +22,27 @@ import { allFinite, assertShapesEqual } from "./tensor_util";
...
+ *
+ * If a tensor is given to tensor, it simply returns it.
+ *
+ * @arg t - An array of numbers, representing a Tensor. Or a Tensor in which
+ *          case tensor() acts as the identity function.
+ * @arg args - An object like this { dtype: "int32", device: "GPU:0" }
+ */
+export function tensor(t: types.TensorLike,
+                       opts?: types.TensorOpts): Tensor {
+  if (t instanceof Tensor) {
```
piscisaureus: Why do you use `instanceof Tensor` here, and `isTensor` elsewhere?

src/dl/math/backends/webgl/select_gpu.ts
ry: https://github.com/propelml/propel/blob/38597c77f9d2bfc0baf2853c5d9aa54b9a78d231/src/dl/math/backends/webgl/shader_compiler.ts#L422-L430

src/tensor_util.ts
ry: Oh, I guess that works. Fixed.

src/tensor.ts
ry: Fixed.

